PR #8, opened 2023-12-07
Big review of the current codebase

Martinsos:


smilebasti:
Hi, is there an eta for these changes to be merged? Would love to use the latest and updated version :)

Martinsos:
Oh no worries, the latest version is on `main`, this is just a very specialized PR which I was just using to do comprehensive review of codebase but never got to it. It is not 22k of changes, it is no changes at all actually, just a placeholder. I will close it so it doesn't confuse people. We are improving OpenSaas all the time and all the changes are pretty quickly landing on `main`, so you are not missing out on anything. The next bigger update will be coming in a couple of weeks, when we switch to Wasp 0.12, as soon as it releases.

PR #14, opened 2024-01-06
new icons, demo app improvement

martinovicdev:
- Added icon for deletion of Task + Lucide Icons pack for other uses 
- Changed behaviour so Task can be deleted at all times 
- Adding Task when pressing Enter key
- Small colour fixes

### DIFF
--- a/app/.prettierrc
+++ b/app/.prettierrc
@@ -1,0 +1,8 @@
+{
+  "trailingComma": "es5",
+  "semi": true,
+  "singleQuote": true,
+  "endOfLine": "lf",
+  "tabWidth": 2,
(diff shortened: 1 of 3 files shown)


vincanger (changes requested):
nice. looks good, @martinovicdev !

can we add a prettier config to use single quotes? that would make the PR easier to read.

martinovicdev:
No problem at all!
I will add one similar to one we have at Wasp repo: https://github.com/wasp-lang/wasp/blob/main/web/.prettierrc

martinovicdev:
Added Prettier with some formatting options hope it is OK now.

vincanger (approved):

PR #26, opened 2024-02-02
Spotted small issues

louisguitton:
I was browsing the demo app and noticed the dark theme had an issue on the Account page for the two dynamic fields.


matijaSos (approved):
Looks great, thanks for the fix! :)

arkhurst (approved):

PR #46, opened 2024-02-15
Updated getting started guide to help Windows users.

martinovicdev:
Created a description that helps Windows users use Wasp.


vincanger (approved):
💅🌭🚀

PR #69, opened 2024-03-05
Playwright tests & CI

vincanger:
Adding some basic playwright tests, as well as a couple github workflows.

Questions:
1. should we include these tests on the main branch, or perhaps only on a feature branch, as well as the `deployed-version` branch (the branch which gets deployed to opensaas.sh)? this way, we can test new features before updating the template on `main`, as well as the app that gets deployed to opensaas.sh without adding a lot of complexity to the template the users initially get. if users DO want to opt in to tests/CI, we can instruct them on how to do so in the docs.
2. should e2e tests for the template test stripe checkout flow? if so, we'd probably need to build a modified feature branch and create a staging environment, because, at the moment, the template on `main` ships with username and password auth only to start, as it doesn't require any ENV VARS and allows the user to get started right away, but it is not compatible with Stripe checkout. Therefore, if we want extensive testing of the template, we should probably go the feature branch/staging environment route...


app/playwright/tests/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+  // Fill input[name="password"]
+  await page.fill('input[name="password"]', user.password || 'password123');
+
+  // Click button:has-text("Sign up")
+  await page.click('button:has-text("Sign up")');
+};
+
+export const createRandomUser = () => {
+  const username = `user${Math.random().toString(36).substring(7)}`;
+  const password = `password${Math.random().toString(36).substring(7)}!`;
```
Zeko369, line 55: Maybe use `crypto.randomUUID()` since it's more obvious what it does, "this looks like some random code"

app/playwright/tests/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+// Create a new Prisma client to interact with DB
+export const prisma = new PrismaClient();
+
+export type User = {
+  id?: number;
+  username: string;
+  password?: string;
+  hasPaid?: boolean;
+  credits?: number;
+};
```
Zeko369, line 12: Import from Prisma?

app/playwright/tests/unpaidUserTests.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  expect(generateScheduleButton).toBeTruthy();
+
+  await Promise.all([
+    loggedInPage.waitForRequest((req) => req.url().includes('operations/generate-gpt-response') && req.method() === 'POST'),
+    loggedInPage.waitForResponse((response) => {
+      // expect the response to be 402 "PAYMENT_REQUIRED"
+      if (response.url() === 'http://localhost:3001/operations/generate-gpt-response' && response.status() === 402) {
+        return true;
+      }
+      return false;
```
Zeko369, line 35: Maybe use just the path since it's easier to run it against non localhost:3001 targets in that case

app/playwright/tests/unpaidUserTests.spec.ts
```diff
@@ -0,0 +1,40 @@
+import { expect } from '@playwright/test';
+import { createLoggedInUserFixture } from './utils';
+
+// Create a new test fixture with an unpaid user and a logged in page
+const test = createLoggedInUserFixture({ hasPaid: false, credits: 0 });
+
+test('Demo app: cannot generate schedule', async ({ loggedInPage }) => {
+  await loggedInPage.waitForURL('http://localhost:3000/demo-app');
```
Zeko369, line 8: I'll just comment here but applies everywhere, use `/demo-app` instead of `http://localhost:3000/demo-app` so it's easier to run against different hosts

app/playwright/tests/paidUserTests.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    ),
+    loggedInPage.waitForResponse((response) => {
+      if (response.url() === 'http://localhost:3001/operations/generate-gpt-response' && response.status() === 200) {
+        return true;
+      }
+      return false;
+    }),
+    // We already started waiting before we perform the click that triggers the API calls. So now we just perform the click
+    generateScheduleButton.click(),
+  ]);
```
Zeko369, line 41: Maybe add a test to see what happened (that the response got shown)

app/playwright/tests/unpaidUserTests.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    loggedInPage.waitForResponse((response) => {
+      // expect the response to be 402 "PAYMENT_REQUIRED"
+      if (response.url() === 'http://localhost:3001/operations/generate-gpt-response' && response.status() === 402) {
+        return true;
+      }
+      return false;
+    }),
+    // We already started waiting before we perform the click that triggers the API calls. So now we just perform the click
+    generateScheduleButton.click(),
+  ]);
```
Zeko369, line 39: Maybe add a test to see what happened (that the error got shown)

app/playwright/tests/landingPageTests.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+    await expect(page).toHaveTitle(/SaaS/);
+  });
+
+  test('get started link', async ({ page }) => {
+    await page.getByRole('link', { name: 'Get started' }).click();
+  });
+
+  test('headings', async ({ page }) => {
+    expect(page.getByRole('heading', { name: 'SaaS' })).toBeTruthy();
+    expect(page.getByRole('heading', { name: 'Features' })).toBeTruthy();
```
Zeko369, line 19: `.toBeVisible()`

app/playwright/tests/landingPageTests.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+    await page.goto('localhost:3000');
+  });
+
+  test('has title', async ({ page }) => {
+    // Expect a title "to contain" a substring.
+    await expect(page).toHaveTitle(/SaaS/);
+  });
+
+  test('get started link', async ({ page }) => {
+    await page.getByRole('link', { name: 'Get started' }).click();
```
Zeko369, line 14: Add `await page.waitForURL('/the-path-this-leads-to')`

Zeko369 (changes requested):

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,54 @@
...
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@v4
+
+      - name: Install Wasp
+        run: curl -sSL https://get.wasp-lang.dev/installer.sh | sh -s
+
+      - name: Docker setup
+        uses: docker/setup-buildx-action@v3
+
```
Martinsos, line 27: You will likely also want to pin down the node version in this node action, so we are explicit about it. Check how we are doing it for `wasp`, we have a step which installs specific version of node.

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,54 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@v4
+
+      - name: Install Wasp
+        run: curl -sSL https://get.wasp-lang.dev/installer.sh | sh -s
```
Martinsos, line 23: So this installs any version of Wasp. If we release new major version of Wasp, but open-saas is not yet on it, this will start breaking. Maybe you should rather pin down the version of Wasp here?

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Docker setup
+        uses: docker/setup-buildx-action@v3
+
+      - name: Cache Node Modules
+        uses: actions/cache@v4
+        with:
+          path: ~/.npm
+          key: node-modules-${{ hashFiles('app/package-lock.json') }}
```
Martinsos, line 32: I would add some more keys to this cache.
I would have: `os`, `package-lock`, `node version`. Because if you add mac in the future, you don't want to pull in node_modules built for ubuntu. Or if we change node version, we don't want to pull in node_modules built for old version. These three should be enough to always pull in the correct node_modules.
We are also now fixing this in Wasp because we are missing it there also, I reached out to @infomiho regarding that. So unless I am wrong, this is something we should do.

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Install Node.js dependencies
```
Martinsos, line 34: Why are we installing deps like this? Shouldn't we use Wasp to install dependencies?

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: |
+          cd app
+          npm ci
+
+      - name: Setup Env Vars not in github secrets
```
Martinsos, line 39: Ok so we just "dumb" copy them right? Aha, so some of them are in Github secrets? Seems to me like you set all of them to example stuff? I am a bit confused how the `cp .env.server.example .env.server` directly reflects "Setup Env Vars not in github secrets", feels like there is a part of story missing here, some assumptions that are not explicitly handled and might change and breaking things in the future.

.github/workflows/retag-commit.yml
```diff
@@ -0,0 +1,28 @@
+name: Retag Commit
```
Martinsos, line 1: So what is our reasoning behind this Action -> it keeps assigning this tag to the last commit on `main`, right?

I guess that makes sense! It means people doing `wasp new` and just copying this template are getting the same version.

WHat happens when we want to update to new Wasp version -> we update the version here in this action, and push it to `main` together with the rest of the changes that update open-saas to this new version, right? Ok yeah, sounds good!

Let's add to README something about this, that explains that this is happening and what we have to do to get open-saas to use new Wasp version.

.github/workflows/retag-commit.yml
```diff
@@ -0,0 +1,28 @@
...
+
+      - name: Delete Old Tag
+        run: |
+          git tag -d wasp-v0.12-template || true 
+          git push origin :refs/tags/wasp-v0.12-template || true
+
+      - name: Add New Tag
+        run: |
+          git tag wasp-v0.12-template
+          git push origin wasp-v0.12-template
```
Martinsos, line 28: You might want to extract this `wasp-v0.12-template` value as an ENV VAR here (whatever Github Actions support, be it on yaml level or env var level), to remove duplication, because I can imagine in the future somebody easily forgetting to update one of these, or missing a letter, or whatever.

app/ci-start-app.js
```diff
@@ -0,0 +1,31 @@
+const cp = require('child_process');
```
Martinsos, line 1: Hm what is this, we would do well with some explanation here. Idealy we would have this here, in `app/`, because it now became part of the template!

app/main.wasp
```diff
@@ -1,6 +1,6 @@
 app OpenSaaS {
   wasp: {
-    version: "^0.12.0"
+    version: "^0.12.2"
```
Martinsos, line 3: Sure, but probably shouldn't be part of this PR, right?

app/package-lock.json
Martinsos, line 1: We didn't have this previously, did we? Why do we have it now?

.github/workflows/e2e-tests.yml
Martinsos: Thinking about this more -> do we want to have package-lock.json for our open-saas template? What are we testing really? Maybe we are testing how it works for a new open-saas app, that just pulled in the latest packages? This does make our tests flaky hm. Ok this goes together with the question of, should the app/ contain package-lock at all.
The problem is, Wasp doesn't yet have good support for package-lock. Even if you have it in app/, the generated .wasp/.../client and .wasp/.../server don't have package-lock's, and you are not really properly guarding yourself. So I wouldn't do it at all probably for now, until we figure that out. Maybe open an issue for it, to look into this and we can discuss it there, but wouldn't bother with it right now.

app/package.json
```diff
@@ -1,5 +1,14 @@
 {
   "name": "opensaas",
+  "scripts": {
+    "example-app:start": "node ci-start-app.js",
+    "example-app:start-db": "npm run example-app:cleanup-db && wasp start db",
+    "example-app:start-app": "npm run example-app:wait-for-db && wasp db migrate-dev && wasp start",
+    "example-app:wait-for-db": "npx wait-port 5432",
+    "example-app:cleanup-db": "(docker container rm $(docker container ls -f name=^wasp-dev-db-OpenSaaS- -q) -f || true) && docker volume rm $(docker volume ls -f name=^wasp-dev-db-OpenSaaS- -q) -f || true",
+    "example-app:playwright": "playwright test",
+    "example-app:playwright:ui": "playwright test --ui"
```
Martinsos, line 10: @vincanger can you give me a bit of reasoning on all this? Are we ok with users having this in their open-saas project once they copy the template? Are they going to be able to understand this -> what is it for, and how were choices here made, how to modify this if needed?

Martinsos (changes requested):
@vincanger I did a review of everything but the Playwright tests themselves -> once we take care of most of the stuff, I can also review those into more details. I was instead focusing on how it is all orchestrated.

Some general stuff:

Normally, you would run CI on every PR and on `main` branch. That way we get CI checks done when for the PR, so we know all is fine before merging into `main`, and then also on `main`, again, just to ensure `main` is all good after the merging and something did go wrong.

This is also how we do it for `wasp` repo.

You mentioned testing Stripe but for that you need a different type of auth, not username & pass, and how you could handle that using different git branches.
In general, I would say that is not the usual approach -> you don't want to get entangled into maintaining multiple branches at the same time -> instead normally you would aim for having these other versions as some kind of example apps in the repo, that you run these tests on. So maybe there would be "examples" dir at the top level of the repo, and you would test this stuff there. Not saying that doing something with branches can't be done, maybe it can and maybe it is potentially interesting, but I wouldn't consider it as a go-to option.

Btw, we are likely to switch open saas to using email&pass very soon because it has a good support now (with 0.12) for running in dev with no setup at all, so maybe that will help a bit for Stripe testing?

As for giving them these e2e tests or not, setup via Github Actions -> my first instinct is to let them have it, by default, since it is nice, does the tests for you, and we give them instructions how to delete it if they don't like it.

ALthough retag-commit is problematic, we don't want them to have that one hm!

If they use this template via `wasp new`, then it is easy, but if they clone the Github repo as a template, then I don't know how we can do this properly.
I am tempted to say we allow using it only via `wasp new`! It gives us more control, we can do what we want here in repo and `wasp new` can delete some files upon download, or download only some files (although they don't get git repo immediately).

.github/workflows/e2e-tests.yml
Martinsos: Ok maybe I had s bit of a braibfart here - it is normal for us with wasp 0.12 to have package-lock actually. But our story about package-locks is indeed still shaky. And this is a template so I wonder if having package-lock in it makes sense, that still stands. Hm. We should think this through a bit.

app/package-lock.json
Martinsos: Hm I take this back, we have it because of Wasp 0.12 which adds it, right? Still, the question is do we want to have in the template -> that is something to consider.

app/playwright/tests/utils.ts
vincanger: password doesn't exist on the user model since v0.12.x uses a new obfuscated auth structure

Martinsos:
@vincanger and I discussed the structure of e2e tests a bit, decided they were too integrated with the Wasp app (they were entangled into its package.json and similar), we agreed to separate them so they are self-contained.

.github/workflows/e2e-tests.yml
vincanger: i believe I managed to do this via this sep:

```sh
      - name: Setup Node.js
        uses: actions/setup-node@v4
        with:
          node-version: '20'

      - name: Install Wasp
        run: curl -sSL https://get.wasp-lang.dev/installer.sh | sh -s

      - name: Docker setup
        uses: docker/setup-buildx-action@v3
 ```

File: .github/workflows/e2e-tests.yml (hunk 10 above)
Comment (vincanger) [reply .github/workflows/e2e-tests.yml:29]: i should probably bump it up to 0.12.4, right?

File: .github/workflows/e2e-tests.yml (hunk 13 above)
Comment (vincanger) [reply .github/workflows/e2e-tests.yml]: ok i renamed this to Setup Dummy Env Vars. they are just placeholders to keep the app from crashing. I will add in the docs that you need to add certain env vars to the github env depending on your tests

File: app/main.wasp (hunk 17 above)
Comment (vincanger) [reply app/main.wasp:3]: :(

File: app/package-lock.json
Comment (vincanger) [reply app/package-lock.json:1]: moved everything to the `e2e-tests` directory as its own npm project

File: app/package.json (hunk 18 above)
Comment (vincanger) [reply app/package.json]: same here

**Comment by vincanger** [comment]
ok @Martinsos -- I made the requested changes. could you review again, please?

File: .github/workflows/e2e-tests.yml (hunk 10 above)
Comment (Martinsos) [reply .github/workflows/e2e-tests.yml]: Yeah and I see you bumped it, great!

What you will likely want to do is update this version number any time you update the Wasp version that open-saas uses. If you forget, I think this will just crash, so that is good, that will remind you. But what you could also do is a add a bit into README.md on what needs to be done when updating the version of Wasp that open-saas uses, and there you can note that it also needs to be updated here. You will know the best if that makes sense or not so I will leave it to you to decide if that is valuable.

File: .github/workflows/e2e-tests.yml [hunk 19]
```diff
@@ -0,0 +1,66 @@
+name: e2e tests
+
+on:
+  push:
+    branches:
+      - main
+      - playwright-tests
```
Martinsos, line 7: Since this is a branch toward `main`, I don't think you need this line with `playwright-tests` here, it is being triggered anyway.

.github/workflows/e2e-tests.yml
Martinsos: Ok, you added OS and node version, which is great. I am just curious, you added them in front of `node-modules` part of the key, why is that? I think it doesn't matter much, but I would naturally put it after, so if I am reading that key somewhere, I first read `node_modules`, aha, these is a cache of node modules, and then come the details of the environment it was cached for (os, node version, package-lock version).

Martinsos: What remaains open is the question of package-lock.json. I wrote bunch a stuff above, said we should discuss -> what is your line of thinking about it?

Martinsos: Oh, another question -> what about e2e-tests? They are also an npm project. They also have their own node_modules, and their package-lock.json and so on. What about them? Do we care about their package-lock.json?

Btw, I don't know how much details you know about how `actions/cache@v4` GH action works, but I remember I was surprised a bit, because it won't update the cache if it managed to pull it. So either if finds it and pulls it in, or it doesn't and then will upload the new cache.
I found this problematic for `waspc`: https://github.com/wasp-lang/wasp/blob/main/.github/workflows/waspc-ci.yaml#L117 -> because I didn't want the whole cache dropped if one dependency changed. I wanted it to be more like the normal behaviour on my dev machine -> restore the last state of npm packages cached on the machine and once done, just update it. But instead I was getting more of "all or nothing" situation. Which for Haskell was a major issue, because if cache gets dropped, it takes 20 minutes to install all the deps! Even though like only 1 dep got added, I would loose thousands of them that were already in the cache. Which is why I did a bit more involved setup there.

You might want to do something similar here, or not, I am not 100% sure yet. But there is also the question of wasp version, I just realized -> let's say everything remains the same, node version, os version, package-lock.json in app, but version of wasp changes. Your cache key will not change, I think? Or maybe it will because package-lock.json will change? I think it is possible it won't change, if new wasp version brought a change in deps of "framework" code. So thinking about it, wasp version should also probably be part of that npm cache.

So there are two things to discuss here:
- how does actions/cache really behave and how we want to use it then
- which part's of this project are contributing to the `./npm` and therefore should be considered while caching it (at the moment I would say user's package.json, e2e package.json, and also wasp's framework code package.json's (which can be tied to wasp version)).

.github/workflows/e2e-tests.yml
Martinsos: So you uncommented this out. Why not delete it?

Aha, but you do have it a bit down there, but now running it in `e2e-test/`. Ok.

.github/workflows/retag-commit.yml
```diff
@@ -0,0 +1,42 @@
+name: Retag Commit
+
+# NOTE: this action is only necessary for development of the Open SaaS template.
```
Martinsos, line 3: Ha yeah, this is the consequence of using this repo as a template, we now have redundant stuff like this.

This is why I would be highly tempted to remove the option to copy this as a template, and just have them pull it through `wasp` CLI (as they can do right now), where we can then do the necessary cleanup as needed. Might be worth opening an issue for this, and following up on it. It will make it easier for both us and them I think.

.github/workflows/retag-commit.yml
```diff
@@ -0,0 +1,42 @@
+name: Retag Commit
+
+# NOTE: this action is only necessary for development of the Open SaaS template. 
+# If you are not contributing to the Open SaaS template itself,
+# and just using it to build your own app, then you can delete this file.
+
+# This action keeps assigning the TAG_NAME to the last commit on main.
```
Martinsos, line 7: Great, but I still think this should be in README.md, under `Contributors` section. When I am doing stuff on a repo, I will be looking into README.md for instructions on what I have to do when I am 1) updating version 2) releasing 3) something else. I probably will never even check out this file otherwise.

Same is for the users, we should probably have a section in the README.md that tells them which parts of this repo they should delete once they copy it. As I mentioned, I think there is a better way to handle this, by not allowing forking it as a template, but in the meantime, this is also possible solution.

.github/workflows/e2e-tests.yml
Martinsos: Ok, so this is basically prepping the app so it can be run.

I am wondering if potentially this should be done as a part of e2e-tests logic though (probably inside their package.json). Not sure yet, I have to read more of the code.

e2e-tests/tests/paidUserTests.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+test('Demo App: add tasks & generate schedule', async ({ loggedInPage }) => {
+  const task1 = 'create presentation on SaaS';
+  const task2 = 'build SaaS app draft';
+
+  await loggedInPage.waitForURL('/demo-app');
+
+  // Fill input id="description" with "create presentation"
+  await loggedInPage.fill('input[id="description"]', task1);
+
+  // Click button:has-text("Add task")
```
Martinsos, line 17: Redundant comment

e2e-tests/tests/paidUserTests.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+const test = createLoggedInUserFixture({ hasPaid: true, credits: 10 });
+
+// test /demo-app page by entering "todo" and clicking add task
+test('Demo App: add tasks & generate schedule', async ({ loggedInPage }) => {
+  const task1 = 'create presentation on SaaS';
+  const task2 = 'build SaaS app draft';
+
+  await loggedInPage.waitForURL('/demo-app');
+
+  // Fill input id="description" with "create presentation"
```
Martinsos, line 14: Redundant comment (I guess it was here for copilot).

e2e-tests/tests/paidUserTests.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  await loggedInPage.click('button:has-text("Add task")');
+
+  // expect to find text in a span element
+  expect(loggedInPage.getByText(task1)).toBeTruthy();
+  expect(loggedInPage.getByText(task2)).toBeTruthy();
+
+  // find a button with text "Generate Schedule" and check it's visible
+  const generateScheduleButton = loggedInPage.getByRole('button', { name: 'Generate Schedule' });
+  expect(generateScheduleButton).toBeTruthy();
```
Martinsos, line 30: So I am not playwright expert but checking for `toBeTruthy` doesn't sound the same as checking it the button is visible. Maybe that is ok, maybe you don't want to check if it is visible, but just that the dom element exists. But then the comment is potentially wrong?

e2e-tests/ci-start-app.js
```diff
@@ -0,0 +1,31 @@
+const cp = require('child_process');
```
Martinsos, line 1: Would be intersting to undresatnd a bit better why this script is needed. Because we want to run these two in parallel? There isn't a simpler way? From what I got we have to handle some special case with Prisma, somebody mentioned that? If so, that should be mentioned here as a motivation. I know you copied this from `wasp`, but that means it is confusing there also, let's improve it here if we can.

e2e-tests/tests/unpaidUserTests.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { expect } from '@playwright/test';
+import { createLoggedInUserFixture } from './utils';
+
+// Create a new test fixture with an unpaid user and a logged in page
```
Martinsos, line 4: Redundant comment.

e2e-tests/tests/paidUserTests.spec.ts
```diff
@@ -0,0 +1,55 @@
+import { expect } from '@playwright/test';
+import { createLoggedInUserFixture } from './utils';
+
+// Create a new test fixture with a paid user and a logged in page
```
Martinsos, line 4: Redundant comment, code says exactly the same thing (which is great, nice code).

e2e-tests/tests/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+  await page.goto('localhost:3000');
+
+  // Click the get started link.
+  await page.getByRole('link', { name: 'Log in' }).click();
+
+  console.log('logging in...', user);
+  await page.waitForURL('http://localhost:3000/login');
+  console.log('url', page.url());
+
+  // Fill input[name="username"]
```
Martinsos, line 23: Bunch of redundant comments in this file.

e2e-tests/tests/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+  await page.fill('input[name="username"]', user.username);
+
+  // Fill input[name="password"]
+  await page.fill('input[name="password"]', user.password || 'password123');
```
Martinsos, line 27: Extract `password123` into a constant `DEFAULT_PASSWORD` and use that in both places.

e2e-tests/tests/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+      await page.waitForURL('/demo-app');
+      const user = await prisma.user.update({
+        where: { username: testUser.username },
+        data: { hasPaid: testUser.hasPaid, credits: testUser.credits },
+      });
+      await use(page);
+      // clean up all that nasty data 🤮
+      await prisma.gptResponse.deleteMany({ where: { userId: user.id } });
+      await prisma.task.deleteMany({ where: { userId: user.id } });
+      await prisma.user.delete({ where: { id: user.id } });
```
Martinsos, line 76: Oh ok so this is not only creating a new user, but also deleting any data that user might already have in the database? That sounds a bit unexpected. ALso, if this user is random, how can there be any data from before?
I think either this is not needed, or it is worth considering doing it in a separate piece of logic.

app/package.json
Martinsos: Ok, we figured this stuff out!

Martinsos (commented):
@vincanger nice job with the changes, this is now quite better, getting there!
I left some more comments.

.github/workflows/retag-commit.yml
vincanger: on the other hand, a tag will not create any issues if they fail to remove it from their repo. I could also add a section in the docs that gives them a command they can copy and paste to remove this file. Even if we remove it via `wasp new` we can't stop people from cloning the repo...

app/package-lock.json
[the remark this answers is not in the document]
vincanger: I had Playwright installed on the app, but then removed it and moved it to the e2etests folder

e2e-tests/tests/utils.ts
```diff
@@ -0,0 +1,78 @@
+import { test as base, type Page } from '@playwright/test';
+import { prisma } from '../../app/src/server/utils';
```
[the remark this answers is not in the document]
vincanger: i like to live dangerously

e2e-tests/package.json
```diff
@@ -0,0 +1,28 @@
...
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@playwright/test": "^1.42.1",
+    "linebyline": "^1.3.0"
+  },
+  "devDependencies": {
+    "@types/node": "^18.0.0",
+    "prisma": "^5.11.0"
```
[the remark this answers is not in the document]
vincanger: i probably forgot to uninstall this

e2e-tests/tests/paidUserTests.spec.ts
vincanger: haha yes. i'll remove them

e2e-tests/tests/utils.ts
vincanger: as I understand it, the user is created, and then tests are ran against it via line 73 `await use(page)` where it pauses and awaits the completion of the tests. After tests run, the user data is deleted.

.github/workflows/e2e-tests.yml
vincanger: I left it there to test the action every time I pushed changes to `playwright-tests`. will delete it before we merge!

e2e-tests/ci-start-app.js
vincanger: paging MR. MIHO on this one, too @infomiho

infomiho: In the original script we are mentioning an issue with the link 🙂 

It boils down to how we check DB connectivity by running Prisma commands and it relies on stdin working properly. This script made it work for us in the CI and none of the fancy parallel npm tooling 🤷🏻‍♂️

e2e-tests/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "version": "1.0.0",
+  "description": "e2e-tests for opensaas.sh",
+  "main": "ci-start-app.js",
+  "directories": {
+    "test": "tests"
+  },
+  "scripts": {
+    "e2e:start": "node ci-start-app.js",
+    "e2e:start-db": "npm run e2e:cleanup-db && cd ../app && wasp start db",
+    "e2e:start-app": "npm run e2e:wait-for-db && cd ../app && wasp db migrate-dev && wasp start",
```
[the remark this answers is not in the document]
vincanger: i don't think this is necessary, since the command `e2e:start-db` is a process being called within `ci-start-app.js`. so db is alwasy cleaned up before starting it.

e2e-tests/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "e2e:wait-for-db": "npx wait-port 5432",
+    "e2e:cleanup-db": "(docker container rm $(docker container ls -f name=^wasp-dev-db-OpenSaaS- -q) -f || true) && docker volume rm $(docker volume ls -f name=^wasp-dev-db-OpenSaaS- -q) -f || true",
```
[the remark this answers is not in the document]
vincanger: leaving comments in a json file is kinda weird, e.g. 

```json
"_comment-on-e2e:cleanup-db": "NOTE: the name of the DB container, e.g. name=^wasp-dev-db-OpenSaaS-, is generated by wasp and will match the name of the app definition in your `main.wasp` file, e.g. `app OpenSaaS { }`",
"e2e:cleanup-db": "(docker container rm $(docker container ls -f name=^wasp-dev-db-OpenSaaS- -q) -f || true) && docker volume rm $(docker volume ls -f name=^wasp-dev-db-OpenSaaS- -q) -f || true",
```

e2e-tests/tests/utils.ts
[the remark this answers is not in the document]
infomiho: Ideally, we don't need to manipulate the DB directly when we are doing e2e tests. Somehow we should be able to perform all the tests steps via user interaction only. If there are some steps that need to be mocked for example, Stripe payments, we should maybe short-circuit it some custom env vars in the app which we set to `true` when running the e2e tests?

That being said, I've seen here for example that they do use direct DB manipulation in tests: https://www.prisma.io/blog/testing-series-4-OVXtDis201

If we need to do it, ideally, we should not import Prisma from the Wasp project, but we would set up Prisma in the `e2e-tests` project:
1. Install Prisma as a dependency
2. Init the `PrismaClient` by referencing the schema file from the Wasp project `npx prisma generate --schema=...`
3. Use the `PrismaClient` by importing it from `@prisma/client`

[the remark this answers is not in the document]
infomiho: I'll try to get this working tomorrow and report back 👍

.github/workflows/e2e-tests.yml
Martinsos: Soudns good! I think you can delete it right now though, because you have a PR now, so it will be run for it anyway, right? Because it is a PR.

.github/workflows/retag-commit.yml
Martinsos: Sure, but this is not the only thing that doesn't make sense for them, there are multiple things that are a mismatch between what we need for this repo, and what the template for them should look like.
As for cloning -> well we can't stop them from cloning the repo, but we can direct them to not do it. If we say in docs they should pull it via `wasp new open-saas`, they will follow that.

app/package-lock.json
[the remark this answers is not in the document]
Martinsos: Yup! Let's make sure though that we don't have these old changes here then.

e2e-tests/tests/utils.ts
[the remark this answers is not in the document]
Martinsos: Yeah I get what you mean by ideally not manipulating the database, you are right, that would be perfect.
That said, sometimes it can be practical to manipulate the database.
And in that case, yeah it would be better if we have local prisma that uses Wasp's schema.
So I agree with everything you said here, sounds right @infomiho ! Interested to see what you come up with.

e2e-tests/package.json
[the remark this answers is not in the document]
Martinsos: Oh yeah I know! But you did fine there I think.

e2e-tests/playwright.config.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    /* Test against branded browsers. */
+    // {
+    //   name: 'Google Chrome',
+    //   use: { ...devices['Desktop Chrome'], channel: 'chrome' },
+    // },
+  ],
+
+  // only run the web server on CI
+  webServer: process.env.CI
```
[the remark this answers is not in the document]
Martinsos: @vincanger you put a :rocket: emoji but I don't see any comment added in the code: am I seeing something wrong? Or maybe you didn't push these changes?

e2e-tests/tests/landingPageTests.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { test, expect } from '@playwright/test';
+import { DOCS_URL } from '../../app/src/shared/constants';
```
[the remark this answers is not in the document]
Martinsos: Yeah let's hardcode it then!

e2e-tests/tests/paidUserTests.spec.ts
Martinsos: Same like in the comment above, you put a :rocket: but I can still see the comment.

Martinsos: Ok my bad, Github took me to the old commit which is why I thought these are not yet done! All good here.

e2e-tests/ci-start-app.js
Martinsos: Ok cool, let's make sure to cover this with the comment then!

e2e-tests/tests/utils.ts
Martinsos: Ah got it! I completely missed the `use(page)` line.

@infomiho I think database deletion is done before running all the e2e tests, but this deletion here is done after every "single" test I guess, that uses this fixture, so that makes sense to me? Keeping it clean for other tests coming after it.

What @infomiho is proposing those is compelling though, doing tests without having to explicitly "doctor" the database. But if that is quite harder / less practical, then I don't mind us doctoring it.

e2e-tests/playwright.config.ts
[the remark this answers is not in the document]
Martinsos: Hm I still dont' quite get it!

Why would port conflicts happen in CI? What does it even mean you run the web server on CI, what is web server here? Why does it help in local development that we do this stuff manually? I think I am missing quite a bit of context / knowledge here, which is why I don't understand this, but we can assume other people reading this might also be missing this context, so it would be great if we enriched the comment to make it understandable for me/them.

e2e-tests/package.json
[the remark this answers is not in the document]
Martinsos: So `ci-start-app.js` script runs two things at the same time, in parallel: `e2e:start-db` and `e2e:start-app`, if I am correct.

What seems like a potential problem to me is that, due to that, first part of `e2e:start-db`, which is db cleanup, and also first part of `e2e:start-app`, which is waiting for db, may run at the very same time, possibly resulting in `wait-for-db` running a tiny bit sooner, enough to detect already running db and saying "ok, great, it is there", only to then have that same database cleaned up, causing trouble down the line, because now migrate-dev is being run but db is not ready yet.

@infomiho I believe this is same/similar as in e2e tests in Wasp, so I am also interested what you think about this, am I correctly identifying this as a potential problem, or I am overlooking something?

infomiho:
@vincanger here's a way you can setup Prisma inside of the `e2e-tests` project and avoid importing stuff from the Wasp app :)

https://github.com/wasp-lang/open-saas/pull/86

.github/workflows/retag-commit.yml
vincanger: ok. sounds good. i'll go ahead and remove the option to use this repo as a template and add info in the readme / docs for prioritizing `wasp new open-saas`

e2e-tests/tests/utils.ts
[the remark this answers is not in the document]
vincanger: merged miho's PR that creates a new Prisma Client in the e2e-tests folder based on the prisma.schema file

e2e-tests/ci-start-app.js
vincanger: this comment is already below starting on line 11

e2e-tests/package.json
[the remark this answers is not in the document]
vincanger: i believe this is covered by the comment in `ci-start-app.js` on line 11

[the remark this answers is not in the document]
Martinsos: @vincanger from what I am understanding, that comment is talking about the specific issue of prisma hanging if not run via this script that we have here, and is basically the reason why we have this script. But it doesn't address what I am talking about here, there are separate concerns, not connected.

e2e-tests/playwright.config.ts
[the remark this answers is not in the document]
Martinsos: Ah ok, now I think this is much clearer!

Ok, so I will try to rephrase this, to see if I got it right.

What we want, is to run a Wasp app, and then when it is running, we want to run playwright, which then interacts with it and tests it that way (e2e tests).

Playwright has this option, `webServer`, which tells it how to run the actual thing you want to test, in our case "Wasp app". So it will first run that, and then go for testing it.

So just to extra clear, it is not a "playwright webserver" in the sense that it is some kind of web server that playwright runs for its purposes to execute tests, it is instead our wasp app that it is executing tests against.

However, we have an issue, where after playwright is done, the whole things hangs for some reason, and that is annoying as you can't easily just do `wasp start` after it, as it will get into conflicts with it (I guess ports are occupied?).

This isn't a problem in CI, because we won't be running anything after it (hopefully? What if we do in the future?), we will just close the machine.

But it is an issue in development. So in development, we do it differently: we do it manually, by running `wasp start` ourselves, and then playwright won't try to run it on its own, because it sees that webserver is already running (so the docs say here https://playwright.dev/docs/test-webserver), so it will just the existing one already.

Ok! It doesn't have to be this long, the explanation, but I think you can improve your comment even a bit more with some of the info from here. Imagine Matija is reading this -> we want him to know, after reading it, the basic gist of what is it about, without having to go learn Playwright, or knowing what "webserver" stands for in this specific context.

Btw, this sucks a bit, that we have different behaviours during local development and during CI, because of this bug. I am tempted to say we should use the same approach both in CI and development -> let's also use `wasp start` in CI then! Aha, but we can't, because `wasp start` has issues in CI, which is why we have that script. Oh man. Ok, comment will be enough for now, until we properly resolve this.

Btw, i am for fixing this bug, it would be enough to just do basically catch that kill signal arrived and then forwarding it to spawned processes, in the `ci-start-..` script. so something like this:

```
// Spawn a child process
const child = spawn('command', ['args']);

// Listen for the parent process to be killed
process.on('SIGINT', () => {
    console.log('Received SIGINT. Exiting and killing child process.');
    child.kill('SIGINT'); // Kill the child process
    process.exit(); // Exit the parent process
});
```

e2e-tests/tests/utils.ts
Martinsos: What did we do here at the end? @vincanger feel free to close this if we are done here.

Martinsos:
@vincanger @infomiho I did another quick round of comments.
@vincanger let's dig down into any remaining open comments and figure out what we need to do to bring them to resolution!

e2e-tests/playwright.config.ts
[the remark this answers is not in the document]
vincanger: ok @Martinsos I played around with this some more and ran into the issue. Playwright sends a SIGKILL signal and it wasn't being passed down to the child processes because we had `const spawnOptions = {detached: true};`. It works when it's set to false, and I'm not sure why we had it set as detached before (maybe @infomiho knows).

The issue we still have though is that the db docker container is still running on port 5432 and if we try to start the db via `wasp start db` we get an error. Not sure what the best solution would be, because we can't clean up the db via a SIGKILL signal...

e2e-tests/tests/utils.ts
vincanger: @Martinsos like you said, the deletions are done after every single test, to keep it clean for other tests coming after it.

Off the top of my hand, I can't imagine how I would test things that need a logged in user without directly accessing the DB. Could you give me an example @infomiho? how could I test adding tasks and generating an AI schedule without logging in a user and generating tasks, etc?

e2e-tests/package.json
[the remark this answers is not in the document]
vincanger: @Martinsos I never ran into an issue with `wasp start` being run even if  a docker container was still live on port 5432, but i tested it as you suggested, by running `e2e:cleanup-db` before starting the node script and it works fine, so to be safe I made the change as you suggested

e2e-tests/ci-start-app-and-db.js
```diff
@@ -0,0 +1,42 @@
...
+ * Running it this way gives us more control over that and avoids it hanging. See https://github.com/wasp-lang/wasp/pull/1218#issuecomment-1599098272.
+ */
+
+const cp = require('child_process');
+const readline = require('linebyline');
+
+function spawn(name, cmd, args, done) {
+  try {
+    const spawnOptions = {
+      detached: false,
```
vincanger, line 12: this was previously set to `true` and because playwright sends a SIGKILL signla, the child processes weren't receiving them.

e2e-tests/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "main": "ci-start-app-and-db.js",
+  "directories": {
+    "test": "tests"
+  },
+  "scripts": {
+    "e2e:start": "npm run e2e:cleanup-db && node ci-start-app-and-db.js",
+    "e2e:start-db": "cd ../app && wasp start db",
+    "e2e:start-app": "npm run e2e:wait-for-db && npm run e2e:prisma-db-setup && cd ../app && wasp db migrate-dev --name test-run && cd ../e2e-tests && npm run e2e:set-db-name && npm run e2e:prisma-db-setup && cd ../app && wasp start",
+    "e2e:wait-for-db": "npx wait-port 5432",
+    "e2e:set-db-name": "chmod +x ./setupDatabaseName.sh && ./setupDatabaseName.sh",
```
vincanger, line 14: i've added this script to make sure we're setting the correct DATABASE_URL when running the github action. When deploying it with the database name given locally, it failed. For whatever reason, the hash is different when running via CI

e2e-tests/ci-start-app-and-db.js
```diff
@@ -0,0 +1,43 @@
...
+    console.error(error);
+  }
+}
+
+// Exit if either child fails
+const cb = (code) => {
+  if (code !== 0) {
+    process.exit(code);
+  }
+};
```
Martinsos, line 40: So regarding the database cleanup: what we could do is have a function `cleanup` and then call it both here, if code !== 0, and also if we receive a sigkill, sigterm, or whatever is the most appropriate signal.

e2e-tests/tests/utils.ts
Martinsos: @vincanger I think the idea was that you just log in with the user, using Playwright?

e2e-tests/package.json
Martinsos: Ah yes, because hash depends on the path to the Wasp project! That way, even if you have two Wasp projects on the disk with the same name, they will not accidentally share the database, because hash of the path on the disk to each project is used to identify that project.

e2e-tests/ci-start-app-and-db.js
vincanger: from what I understand, the callback will not be triggered with SIGKILL, it's just abruptly terminated

Martinsos:
> i've gotten the tests to run in the same manner locally and within the CI pipeline (via playwright's `webServer` config), the only remaining issue is that the DB does not close on port 5432 and there is no way to run a callback/cleanup function to shut it down as playwright sends SIGKILL and and not SIGTERM or some other signal that would allow us to graciously stop the db.
> 
> Any suggestions @infomiho @Martinsos on how to deal with this issue? At the moment, this causes the problem that, when a user wants to start their db with `wast db start` it is already running on port 5432 after tests have run...

Ok I completely forgot that SIGKILL can't be handled! And now I learned that Playwright sends SIGKILL immediately, instead of sending SIGTERM first. And they have this issue that is describing exactly the problem we also have: https://github.com/microsoft/playwright/issues/18209 .

Seems like they are working on a fix, but nothing yet.

They do suggest using a workaround with GlobalTeardown (https://github.com/microsoft/playwright/issues/18209#issuecomment-1285645202), so maybe that is something we should investigate? I saw somebody mentioning that is too aggressive for their case, but might be good for our case?

Martinsos (commented):
Left some more comments!

e2e-tests/playwright.config.ts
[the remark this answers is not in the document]
vincanger: so martin and I tried this out, and because of DB migration issues and errors when cleaning up the DB, we decided to instruct users to start the wasp app manually before running the tests locally :)

e2e-tests/ci-start-app-and-db.js
vincanger: in the end, we decided to let the user start the app manually before running tests locally to work around this issue

e2e-tests/tests/utils.ts
vincanger: in the end, we are not manipulating the db, and rather logging the user in through the app via playwright as @infomiho suggested

e2e-tests/package.json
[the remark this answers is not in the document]
vincanger: through all the testing, this change is working great.

.github/workflows/e2e-tests.yml
vincanger: ok @Martinsos so I tried to address the caching questions by leaving the app cache in the global cache directory (~/.npm) and then putting the e2e-tests modules in a specific dir and explicitly checking for a cache hit on those before installing the necessary deps

vincanger:
okie dokie @Martinsos ready for another review. here's a rundown:

- improved caching of node modules across all projecs
- set up a local test workflow and a CI test workflow
- both workflows run Stripe CLI in the background and we log in the user and pay through stripe as part of the tests to avoid manipulating the db directly through prisma

app/src/server/actions.ts
```diff
@@ -48,22 +48,28 @@ export const stripePayment: StripePayment<string, StripePaymentResult> = async (
...
 
-  let customer: Stripe.Customer;
-  let session: Stripe.Checkout.Session;
+  let customer: Stripe.Customer | undefined;
+  let session: Stripe.Checkout.Session | undefined;
   try {
     customer = await fetchStripeCustomer(userEmail);
+    if (!customer) {
+      throw new HttpError(500, 'Error fetching customer');
+    }
```
vincanger, line 57: i added better error handling here in case the user forgets to add the STRIPE_PRODUCT env vars to either their app or the github action secrets

e2e-tests/package.json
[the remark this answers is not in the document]
Martinsos: Ok, let's keep it then!

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: Setup Dummy Env Vars 
+        run: |
+          cd app
+          cp .env.server.example .env.server
+
+      - name: Cache App Node Modules
+        uses: actions/cache@v4
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-wasp${{ env.WASP_VERSION }}-node${{ steps.setup-node.outputs.node-version }}-node-modules-${{ hashFiles('app/package-lock.json') }}
```
Martinsos, line 48: Ok, so seems to me like you have keys for all the right stuff!
I would still advise having `node-modules-` as a prefix, and not having it as the last thing, because it makes more sense to me to start with "what" and then the details of it, but not a biggie.

Martinsos: One thing though: let's say you add something to e2e tests, their package-lock.json changes. You didn't add anything to `app`.
What happens now, is that `.npm` will remain as it was before, but node_modules for e2e tests will change. You will save new packages to `.npm`. But will have new cache for e2e tests `node_modules`.
Maybe ok, but a bit weird. Then I would say let's cache local node_modules on both sides and that is it, not cache `.npm` for one project and local `node_modules` for another.

What I would probably go for is caching this global `.npm` instead of two local `node_modules`. What this means is having just one cache action, which is the first one you have, and then also adding `hashFiles('e2e-tests/package-lock.json')` to its key. And that is it then, you don't need the other action ("Cache E2E Tests Node Modules").

Btw you will likely want to also add
```
          restore-keys: |
            node-modules-${{ runner.os }}-
```
to that action, so that cache is not dropped on any change, but is instead just updated. I wrote this assuming you have `node-modules` as a prefix of your cache key.
This is connected to what I wrote in the other comment previously about cache behaving a bit weird regarding dropping / updating itself, I recommend digging a bit into that to understand what is happening.

.github/workflows/e2e-tests.yml
Martinsos: Makes sense, still feels a bit short to me, I described though in another comment what more could be potentially done!

e2e-tests/.env
```diff
@@ -0,0 +1,3 @@
+# Dev DB connection URL specific for the Open SaaS project
+# Needed for direct DB manipulation via Prisma
+DATABASE_URL=postgresql://postgresWaspDevUser:postgresWaspDevPass@localhost:5432/OpenSaaS-dfca0a13c1
```
Martinsos, line 3: This is specific to the path at which project is though, right? I think I remember you gave up on this approach where it is hardcoded, maybe this is a leftover?

e2e-tests/README.md
```diff
@@ -0,0 +1,3 @@
+These are e2e tests.
+
+(TODO: Make comment on installation, setup (env vars), running, etc.)
```
Martinsos, line 3: Let's do this now?

e2e-tests/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "e2e-tests",
+  "version": "1.0.0",
+  "description": "e2e-tests for opensaas.sh",
+  "main": "ci-start-app-and-db.js",
+  "directories": {
+    "test": "tests"
+  },
+  "scripts": {
```
Martinsos, line 9: So this is quite complicated all together, but it is all well enough named, commented on, so I think it works! It is a bit tempting to maybe spend more time on organizing it / simplifying it, but the effort would not be worth the result at this moment. So nice job!

e2e-tests/playwright.config.ts
```diff
@@ -0,0 +1,47 @@
...
+      name: 'chromium',
+      use: { ...devices['Desktop Chrome'] },
+    },
+  ],
+
+  /**
+   * There seems to be a bug that keeps the webserver open after running tests locally https://github.com/microsoft/playwright/issues/11907
+   * causing errors when trying to run `wasp start` afterwards. To avoid this, we can run the webserver only on CI.
+   * For local development, we start the app manually with `wasp db start` and `wasp start` and then `npm run local:e2e:start`.
+   */
```
Martinsos, line 39:
```suggestion
  /**
   * `webServer` field tells Playwright how to run the app (webServer) it tests.
   * It seems however that there is a bug in Playwright that keeps the web server open after running tests locally https://github.com/microsoft/playwright/issues/11907,
   * causing errors when trying to run `wasp start` afterwards. To avoid this, we let Playwright run web server only in CI (where this is not a problem because after tests we don't do anything else).
   * For local development, where this does pose a nuisance, we start the app / web server manually with `wasp db start` and `wasp start` and then start tests with `npm run local:e2e:start`.
   */
```

Martinsos: @vincanger and please remind me, why don't we do same "manual" approach in CI? Ok, we need `npm run ci:e2e:start` because of Prisma output yeah.
But why don't we have same process, both for CI and locally, where we just run `npm run ci:e2e:start` manually, and then run tests after that? That has issues locally?

e2e-tests/setupDatabaseName.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
Martinsos, line 1: @vincanger can you please remind me why we need this? I know we need to delete the databases that are left over running. But why do we need to set up the database name? Doesn't `wasp start db` take care of it, normally we don't have to set db name in .env?
Or this is because somebody else is connecting to the db and we need to get that name? Who?

e2e-tests/setupDatabaseName.sh
```diff
@@ -0,0 +1,38 @@
...
+    exit 1
+fi
+
+# Step 2: Extract the database connection details
+# Assuming the standard Postgres environment variables are used within the Docker container
+DB_NAME=$(docker inspect --format '{{range .Config.Env}}{{println .}}{{end}}' $CONTAINER_ID | grep POSTGRES_DB= | cut -d'=' -f2)
+
+# Construct the Postgres URL. We assume the host is localhost since it's a local Docker environment
+# and the default Postgres port is 5432. Adjust if necessary.
+DATABASE_URL="postgresql://postgresWaspDevUser:postgresWaspDevPass@localhost:5432/${DB_NAME}"
```
Martinsos, line 17: Right, we also assume the username and pass to match those provided by `wasp db start`.

Martinsos (commented):
Left another round of comments! We are getting close now!

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,109 @@
...
+      # In order for the app to run we need to set env vars even if they aren't used
+      # this step sets mock env vars to pass the validation steps so the app can run 
+      # in the CI environment. For vars that are actually used in tests, we set them in
+      # the GitHub secrets settings and access them in a step below.
+      - name: Setup Dummy Env Vars 
+        run: |
+          cd app
+          cp .env.server.example .env.server
+
+      - name: Cache App Node Modules
```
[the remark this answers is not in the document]
vincanger: Nice. That makes more sense to me now. Thanks :)

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,109 @@
...
+        uses: actions/cache@v4
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-wasp${{ env.WASP_VERSION }}-node${{ steps.setup-node.outputs.node-version }}-node-modules-${{ hashFiles('app/package-lock.json') }}
+
+      - name: Cache E2E Tests Node Modules
+        id: cache-e2e-tests
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/e2e-tests
```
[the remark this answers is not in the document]
vincanger: yeah i was just taking a guess

e2e-tests/.env
vincanger: yep. removed.

e2e-tests/setupDatabaseName.sh
vincanger: It's a leftover from earlier attempts. I've removed it now

app/src/server/payments/stripeUtils.ts
```diff
@@ -9,19 +9,23 @@ const DOMAIN = process.env.WASP_WEB_CLIENT_URL || 'http://localhost:3000';
...
+      customer = await stripe.customers.create({
+        email: customerEmail,
+      });
+    } else {
+      console.log('using existing customer');
+      customer = stripeCustomers.data[0];
+    }
+    return customer;
+  } catch (error: any) {
+    console.error(error.message)
```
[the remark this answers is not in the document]
Martinsos: Ok, if you are doing try/catch, you should certainly rethrow the error.
But now you are rethrowing it and not doing console.error -> that is really the same as not having try/catch at all! Because if onlyt thing try/catch does it retrhow an error, then it really doesn't have any effect.
But, you actually do something -> you turn it into Http 500 error. However, what Wasp server does by default, is turn any error that is not Http error into Http 500 error once it reached the top level of server (this is our default error middleware)! So there is no need for you to do this, and you can really just drop the whole try/catch modification.
Unless you want to keep console.error there, but you removed it. But I wonder why you removed it, you said you found it useful? I did wonder why since I think errors would be printed anyway, but maybe I was wrong about that?

.github/workflows/e2e-tests.yml
Martinsos: Yup, exactly! It is quite convoluted, how Github cache action works, and I was confused by at the beginning, but then learned about this.

e2e-tests/README.md
```diff
@@ -0,0 +1,42 @@
+# Open SaaS e2e Tests with Playwright
+
+These are e2e tests that are written using [Playwright](https://playwright.dev/) for the Open SaaS project. 
+
+They not only serve as tests for development of the Open SaaS project, but also as reference examples for how you can implement tests for the app you build with Open SaaS as a template, if you choose to do so.
+
+## Running the tests
+### Locally
```
Martinsos, line 8: Maybe a silly question, but when I run e2e tsts locally, what happens to the database? Local dev database is used, right? So if I was doing development locally, that same database is just used in tests? Meaning tests start with unknown state of the db + once they are done leave some data in it? How problematic is that, if that is happening?
From my experience, normally we will want to separate local dev database from the database used for e2e tests.

@infomiho your thoughts?

Ooooh and idea btw: we could expand `wasp db start` command to be like `wasp db start my-db-2` or `wasp db start my-e2e-db` -> you can give it a db name as an argument, and it will use that to be able to start a different database. So you could do `wasp db start my-dev-db` or `wasp db start my-e2e-db` and it will treat those as two separate databases. Geniussssss :D!

e2e-tests/README.md
Martinsos, line 1: Nice README all together, coherent, to the point, and has all the important information!

e2e-tests/setupDatabaseName.sh
Martinsos: The whole setupDatabaseName.sh is a leftover? If so, sweet, because it was complex and removing it is awesome!

But, I still see the file here?

Martinsos (commented):
Did another round!

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,109 @@
...
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          STRIPE_KEY: ${{ secrets.STRIPE_KEY }}
+          STRIPE_WEBHOOK_SECRET: ${{ secrets.STRIPE_WEBHOOK_SECRET }}
+          HOBBY_SUBSCRIPTION_PRICE_ID: ${{ secrets.HOBBY_SUBSCRIPTION_PRICE_ID }}
+          PRO_SUBSCRIPTION_PRICE_ID: ${{ secrets.PRO_SUBSCRIPTION_PRICE_ID }}
+          CREDITS_PRICE_ID: ${{ secrets.CREDITS_PRICE_ID }}
+          SKIP_EMAIL_VERIFICATION_IN_DEV: true
+        run: |
+          cd e2e-tests
+          npm run e2e:playwright:debug
```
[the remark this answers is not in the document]
vincanger: ok yeah that was something i just copied over from the wasp headless tests. makes sense! I've changed it now :)

e2e-tests/setupDatabaseName.sh
vincanger: whoops. removed!

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,107 @@
...
+          cd app
+          cp .env.server.example .env.server
+
+      - name: Cache global node modules
+        uses: actions/cache@v4
+        with:
+          path: ~/.npm
+          key: node-modules-${{ hashFiles('app/package-lock.json') }}-${{ hashFiles('e2e-tests/package-lock.json') }}-${{ runner.os }}-wasp${{ env.WASP_VERSION }}-node${{ steps.setup-node.outputs.node-version }}
+          restore-keys: |
+            node-modules-${{ runner.os }}-
```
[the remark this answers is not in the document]
vincanger: ok I wasn't aware of that, I thought you suggested it that way because it would match any combination of keys, but it makes sense that it is looking for a substring of that key.

app/src/server/payments/stripeUtils.ts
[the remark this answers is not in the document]
vincanger: Ok interesting. What was happening was that I was getting a 500 error with a generic message. By logging it to the console I was getting a more descriptive error and was able to pinpoint what was actually the issue. I assumed by throwing the error this way, I would thus get the more descriptive error rather than the generic one.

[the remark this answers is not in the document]
vincanger: I tested it again and I think the reason is because the error was getting "swallowed" by the headless tests running in the CI. So, because no error was propagating up to the browser, I wasn't seeing it unless I logged it to the console with console.error.

e2e-tests/README.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+In another terminal, run the local e2e tests:
+```shell
+cd e2e-tests && npm run local:e2e:start
+```
+
+This will start the tests in Playwright's UI mode, which will allow you to see the tests running in an interactive browser environment.
+
+When the tests finish, you can kill the Stripe CLI process that was started by the tests:
```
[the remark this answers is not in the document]
vincanger: yep, you're right. by adding `npm run local:e2e:cleanup-stripe` to the end of the local run tests script, it runs automatically when cancelling the test process with ctrl + c

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,107 @@
...
+        env:
+          # The e2e tests are testing parts of the app that need certain env vars, so we need to access them here.
+          # These secretes can be set in your GitHub repo settings, e.g. https://github.com/<account>/<repo>/settings/secrets/actions
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          STRIPE_KEY: ${{ secrets.STRIPE_KEY }}
+          STRIPE_WEBHOOK_SECRET: ${{ secrets.STRIPE_WEBHOOK_SECRET }}
+          HOBBY_SUBSCRIPTION_PRICE_ID: ${{ secrets.HOBBY_SUBSCRIPTION_PRICE_ID }}
+          PRO_SUBSCRIPTION_PRICE_ID: ${{ secrets.PRO_SUBSCRIPTION_PRICE_ID }}
+          CREDITS_PRICE_ID: ${{ secrets.CREDITS_PRICE_ID }}
+          SKIP_EMAIL_VERIFICATION_IN_DEV: true
```
infomiho, line 104: We set this env var in `.env.server.example` as well, FYI

e2e-tests/README.md
infomiho: In Rails for example, they recognise that `test` is separate env from dev and production. By default they include a `test` env for the database which implies that there is a `test` database which is used in the tests. 

https://guides.rubyonrails.org/testing.html#the-test-database

So, the idea of having a separate `test` DB which has its own "mock data" that gets loaded for tests is a great idea! For now, with our current setup, we only recognise dev and production envs - so not good for locally running e2e tests. It's okay for the CI setup though.

e2e-tests/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "ci:e2e:start-db": "cd ../app && wasp start db",
+    "ci:e2e:start-app": "npm run ci:e2e:wait-for-db && cd ../app && wasp db migrate-dev && cd ../e2e-tests && cd ../app && wasp start",
+    "ci:e2e:wait-for-db": "npx wait-port 5432",
+    "_comment-on-ci:e2e:cleanup-db": "NOTE: the name of the DB container, e.g. name=^wasp-dev-db-OpenSaaS-, is generated by wasp and will match the name of the app definition in your `main.wasp` file, e.g. `app OpenSaaS { }`",
+    "ci:e2e:cleanup-db": "(docker container rm $(docker container ls -f name=^wasp-dev-db-OpenSaaS- -q) -f || true) && docker volume rm $(docker volume ls -f name=^wasp-dev-db-OpenSaaS- -q) -f || true",
+    "e2e:playwright": "DEBUG=pw:webserver npx playwright test",
+    "_comment-on-local:e2e:cleanup-stripe": "NOTE: because we are running the stripe webhook listener in the background, we want to make sure we kill the previous processes before starting a new one.",
+    "local:e2e:cleanup-stripe": "PID=$(ps -ef | grep 'stripe listen' | grep -v grep | awk '{print $2}') || true && kill -9 $PID || true",
+    "local:e2e:start-stripe": "stripe listen --forward-to localhost:3001/stripe-webhook &",
```
infomiho, line 19: Did we mention in the README that you need to install the Stripe CLI locally to get the e2e tests working?

infomiho:
Good job 👍 

I'd maybe add a note in the README about how to install Stripe CLI locally if you want to run the e2e tests locally.

.github/workflows/e2e-tests.yml
vincanger: I will leave it there in case the user changes their local env so that CI tests don't fail

e2e-tests/package.json
vincanger: @infomiho no we don't but I could add it, although I can't imagine someone running the tests without getting stripe working first (it's outlined in the docs)

.github/workflows/e2e-tests.yml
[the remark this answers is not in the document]
Martinsos: Actually it is looking for a prefix I think! So this way it matches the prefix.

e2e-tests/README.md
Martinsos: Nice @infomiho : I adde dwhat you wrote to the issue I linked above that I created for tracking this idea.

.github/workflows/retag-commit.yml
Martinsos: @vincanger you never added anything about this to some top level README, right?
I would certainly do it, as it can be quite surprising that anything that goes to `main` is basically instantly published, that should be explicitly stated.

Martinsos (approved):
@vincanger we want a long way here, but I think we are good now! Great job, we fixed a lot of stuff!

I approved but, but pls check this last comment: https://github.com/wasp-lang/open-saas/pull/69/files#r1572247170 .

Also, now that we improved stuff here, we should take that and apply the improvements on e2e tests we have in `wasp` repo. You could coordinate with @infomiho around that. I recommend doing that as soon as possible, because otherwise you will forget a lot about what we did here and it will become quite harder.

PR #71, opened 2024-03-06
update conditional classnames using clsx

vincanger:
added clsx wherever conditional classNames with template literals were.


martinovicdev:
I suggest using tailwind-merge as well, in order to reduce class conflict. You can create an util function like this:

```import { clsx, type ClassValue } from 'clsx';
import { twMerge } from 'tailwind-merge';

export function cn(...inputs: ClassValue[]) {
  return twMerge(clsx(inputs));
}
```
and use it instead of cslx in your changed code.

vincanger:
yea I thought about it, but at the moment we don't have any reusable components that take classValue inputs, so is it necessary? the user can then decided if they want to implement it themselves then. or am I thinking about this incorrectly?

martinovicdev:
Tailwind Merge is useful for styling conflicts in conditionals as well. This example where you have multiple classes (text, padding) affecting the same property seems realistic to me: 

```
          className={clsx('flex items-center gap-1 text-sm font-medium text-red-500 py-3 px-2', {
            'text-meta-3 gap-2 p-3 text-green-200': isDeltaPositive,
            'text-meta-5 p-4 text-black': !isDeltaPositive,
          })}
```

Without Tailwind merge you'd have conflicts and incorrect style in the app.

vincanger:
so with twMerge this would be resolved? hmmm

vincanger:
ok @martinovicdev , you've convinced me to let you take this PR over (it seems you know more about this than I do) :) do you mind if I assign you and you make the transition to `cn()` where appropriate? it should be pretty straightforward

martinovicdev:
> so with twMerge this would be resolved? hmmm

Exactly, it would clean up the classes correctly and apply the correct ones. For example, if we have isDeltaPositive as true, we'd have only p-4 and text-black applied and text-red-500 py-3 and px-2 removed, so there are no conflicts.

martinovicdev:
> ok @martinovicdev , you've convinced me to let you take this PR over (it seems you know more about this than I do) :) do you mind if I assign you and you make the transition to `cn()` where appropriate? it should be pretty straightforward

Sure, not an issue

martinovicdev:
This should be OK now.

PR #72, opened 2024-03-06
Cleanup scripts for running app

Zeko369:


vincanger:
nice one, thanks @Zeko369

PR #74, opened 2024-03-06
Update .env.server.example

sunsetsobserver:
The docs at https://docs.opensaas.sh/guides/email-sending/ claim that:

"In the .env.server file, we’ve set the SEND_EMAILS_IN_DEVELOPMENT env variable to true. This means that emails will be sent in development mode.

This is useful for testing, but you can turn it off by setting it to false, and the emails will be logged to the console instead.".

However, this isn't in code.


sunsetsobserver:
Right! Thank you for considering it. Closing it here then.

Martinsos:
@vincanger however now with 0.12, I think this changed? New docs should shed more light https://wasp-lang.dev/docs/auth/email, but I know that we worked on making it possible to go with email&pass from the very start with no setup, and I think we might have even removed that ENV var, I am not 100% sure though. But we have not this `Dummy` emailSender and stuff.

@infomiho can help with this possibly.

PR #86, opened 2024-03-25
Setup tests with local Prisma

infomiho:
Instead of importing `Prisma` from the Wasp project, setup Prisma in the tests project so you can still manipulate the DB during the e2e tests, but keep things a bit cleaner.


e2e-tests/package.json
```diff
@@ -13,15 +13,18 @@
     "e2e:wait-for-db": "npx wait-port 5432",
     "_comment-on-e2e:cleanup-db": "NOTE: the name of the DB container, e.g. name=^wasp-dev-db-OpenSaaS-, is generated by wasp and will match the name of the app definition in your `main.wasp` file, e.g. `app OpenSaaS { }`",
     "e2e:cleanup-db": "(docker container rm $(docker container ls -f name=^wasp-dev-db-OpenSaaS- -q) -f || true) && docker volume rm $(docker volume ls -f name=^wasp-dev-db-OpenSaaS- -q) -f || true",
-    "e2e:playwright": "playwright test",
-    "e2e:playwright:ui": "playwright test --ui",
-    "e2e:playwright:debug": "DEBUG=pw:webserver npx playwright test"
+    "e2e:playwright:db-setup": "mkdir -p prisma && cp ../app/.wasp/out/db/schema.prisma prisma/schema.prisma && prisma generate",
```
infomiho, line 16: I've had to go for the copying of the schema file instead of referencing it with `--schema ../app/.wasp/out/db/schema.prisma` because it would generate the `@prisma/client` in the app project and not in the `e2e-tests` project!

e2e-tests/tests/utils.ts
```diff
@@ -1,6 +1,8 @@
 import { test as base, type Page } from '@playwright/test';
-import { prisma } from '../../app/src/server/utils';
 import { randomUUID } from 'crypto';
+import { PrismaClient } from '@prisma/client';
+
+const prisma = new PrismaClient();
```
infomiho, line 5: Using the "local" Prisma instead of importing it from the app :)

PR #106, opened 2024-04-23
Small updates to getting-started.md

Martinsos:
A couple of small fixes, let's see which ones we want to keep.


vincanger:
> A couple of small fixes, let's see which ones we want to keep.

I like them all. Nice!

PR #113, opened 2024-04-29
Fix: Updating subscription status check on generateGptResponse

TheSirLan:
- Change to ensure that deleted and past_due subscriptions are not allowed to access the GPT API without credits.


Martinsos:
Thanks for the pull request! @vincanger is absent at the moment but will check it soon when he is back!

vincanger (approved):
awesome! thanks for the contribution :)

PR #132, opened 2024-05-21
Removed top level .prettierrc.

Martinsos:
While looking at top-level .gitignore, I realized there is also a top-level .prettierrc.
Is this one also a leftover? It seemed a bit alone there, out of context. We could leave it in theory, but since there is one already in `app/`, maybe best if we remove this one?

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,0 @@
-{
-  "semi": true,
-  "singleQuote": true,
-}


vincanger:
@infomiho added this top-level prettier config file (I overlooked the fact that there are two now). maybe he can explain why he put it there, @Martinsos

infomiho:
Since we have code in the `blog` folder as well, the top-level made sense to me 🤷 it helps with formatting when different people work on the project

Martinsos:
> Since we have code in the `blog` folder as well, the top-level made sense to me 🤷 it helps with formatting when different people work on the project

Sure, I just thought it would make more sense if each project had its own: app/, blog/ and so on? Because app/ already has its own anyway. Sounds like we should either have one big one or one per project? Or I guess we have this one global one that has some basic stuff and then each project can configure more? Ok, that makes sense I guess! I will close this PR, I thought this is an unintentional leftover.

PR #134, opened 2024-05-21
main.wasp polish

Martinsos:
I did a bunch of small improvements to main.wasp and grouped them all in this one PR!

### DIFF
--- a/app/main.wasp
+++ b/app/main.wasp
@@ -8,10 +8,10 @@ app OpenSaaS {
   head: [
         "<meta property='og:type' content='website' />",
         "<meta property='og:title' content='My Open SaaS App' />",
-        "<meta property='og:url' content='https://opensaas.sh' />", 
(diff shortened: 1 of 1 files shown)


app/main.wasp
```diff
@@ -51,98 +58,114 @@ app OpenSaaS {
...
 }
 
-/* 💽 Wasp defines DB entities via Prisma Database Models: 
+/* 💽 Wasp defines DB entities via Prisma Database Models:
  * https://wasp-lang.dev/docs/data-model/entities
  */
 
 entity User {=psl
   id                        Int             @id @default(autoincrement())
+  createdAt                 DateTime        @default(now())
```
Martinsos, line 94: In all entities, I grouped fields and ordered them so that most generic ones go first (id, createdAt), then go important relations like `user` + `userId` and similar, and then go business logic fields.

vincanger (approved):
I like!

PR #135, opened 2024-05-21
Replaced string routes with typed routes.

Martinsos:


Martinsos:
This way, if for example you delete file upload stuff, including the page and its route, you will get warned by TS that the route doesn't exist anymore.

vincanger (approved):
TIL Wasp has this routes function :)

Martinsos:
> TIL Wasp has this routes function :)

Yes @infomiho added it relatively recently! It is great, helps you realize that route doesn't exist anymore.

PR #136, opened 2024-05-22
Remove sticky positioning from AppNavBar

vblazenka:
The goal of this PR is to remove unnecessary  `sticky` class from `app/src/client/components/AppNavBar.tsx` due to already existing `absolute` positioning being set.


vincanger (approved):

Martinsos:
Awesome, thanks @vblazenka !

PR #138, opened 2024-05-22
Added warning to CONTRIBUTING.md to not commit migrations/ dir.

Martinsos:


vincanger:
@Martinsos can I close this without merging? We've added `template/app/migrations` to `.gitignore` so I think this message is redundant.

Martinsos:
@vincanger yes good catch, closed it!

PR #154, opened 2024-05-25
All in dirs

### LINKED ISSUE #133: Docs and demo app being on a separate branch is weird
The fact that docs for the template, and a demo app (which is template + some small modifications) are on a separate branch (`deployed-version`) from the actual template (`main`) is weird.

When updating the template, one would normally also want to update the docs, and maybe also the demo app. We should be able to update all of these in one PR.

What we are actually doing here right now is using multiple branches to have multiple projects in the same repo, but that is not how this should be done, we should use directories for that.

One reason for doing this before was that we allowed people to create a new repo by using GitHub's "template repo" functionality. We gave up on that, so that is not a valid reason any more.

One could also argue that having a demo app and docs on a separate branch allows for easy updating of them with new changes from the template (you merge in new stuff from the template when it updates). That does however also mean that those changes can't be done in the same PR in which they are done for the template, which is not great. But this is a better argument, and the question is, how are we going to be updating demo app and docs?

So, let's try to figure this out!

## Current approach

Ok, so problems with the current approach:
1. You can't update docs or demo in the same PR where you are updating the template.
2. You have to do the changes twice -> you have to make one PR for updating the template, then another PR where you do those same changes to the demo app and docs.
3. It is just weird.

I wonder, how do we now make sure that demo app and docs are up to date? If somebody does a PR to `deployed-version` branch, does it get deployed automatically, both app and docs? I know we ask users to make changes for docs against `deployed-version`, but who makes sure that the demo app gets updated with those changes and re-deployed?

Does the fact that demo app is on separate branch help with updating it when template changes? Is it always one big fat commit on top of commits from template so you just keep rebasing that one, or merging into it?

## All in dirs

What if we put stuff into directories?

So at the top level we would have `template/` and `demo-app/` (or `open-saas-sh`).

Then, one can make changes to everything in a single PR.

But, the bad thing is, they should make changes to everything in a single PR. If we did use some branch tricks to help us update the demo app based on changes in the template, we can't use that anymore, at least not as easily.

So how do we easily update demo-app and docs, when template changes? Maybe we could have them implemented in some smart way: so instead of having them be duplicates of template and its docs, we would represent them as patch files, and would have a script that applies patch files to the template to produce demo app and its docs? Is there some existing tooling or something that would help with this? Sounds a bit too much?

## Separate repo

So instead of having multiple branches, we could have demo app in a separate repo. I am not sure we get much with this though. Kind of makes sense, because now it is an actual another app that is made from open-saas template, but other from it semantically being "better", I don't see any other benefits, it has all the same problems as the approach with multiple branches. It does feel a bit better though, less/not wrong.

## What now

Ok, this is a problem statement + some potential solutions: let's discuss!

Martinsos:
Fixes #133 .

I was curious about this whole situation with putting demo app next to the template and what are the diffs really, so I implemented that here.

I checked out `deployed-version` branch and put it into `opensaas-sh` directory.

I also wrote a script (dope.sh) that generates a diff between the `template` and `opensaas-sh` dir, and some helper tools.

Then I went manually through those diffs and tried figured out with help of @vincanger  which of them contains something important vs can be just dropped due to improvements in the template and would be just copied from there anyway to be the latest.

I have done all this only for app/, I didn't touch blog/ or e2e-tests/ .

I also removed retagging, because these changes will require new Wasp version (0.14) in order to pull from the `template` subdir, so either we keep this on some branch like `develop` on the side, or we disable auto retagging. I would rather maybe we keep developing stuff on `main` with retagging disabled though, so we can keep adding other stuff on top of this. And then later we can re-enable auto-tagging and also see how we can automate more stuff in the CI (publishing demo app, releasing docs).

I didn't have anything to edit in OpenSaas docs it seems, is it possible I missed something or is that really so?

I am not 100% sure that approach with diffs will be the best, but I would suggest we give it a try -> if we don't like it, we can generate the `app/` from the diffs one final time and create another repo just for opensaas-sh based on it. In any case, it might be useful in some other applications, so it is not a waste.

TODO:
- [X] Update docs if needed
- [X] Consider cleaning up env situation in app/.gitignore .
- [x] Figure out exactly how opensaas.sh is deployed and write down instructions in its README.md. I wrote basic instructions, but they are lacking details / exact instructions. We need to add exact instructions for deploying to Netlify, and we also need to double check if some client env vars (stripe?) need to be provided to `wasp deploy fly deploy` or not. Vince can help me figure this out.
- [x] Add README.md to `template` and also add one to `template/app`. Maybe also to `tempalte/blog` or `template/e2e-tests`?
- [x] Go with Vince through diff-notes.md, really figure out what is important from the diffs, and then keep only those diffs, polish them, and actually commit them into the source tree under `opensaas-sh/app_diffs`, while removing `opensaas-sh/app`.
- [x] What about ignoring `opensaas-sh/app` in .gitignore? That messes up our diffing, so if we do it, we need to somehow make our diff/patch scripts work around it (probably by them editing .gitignore temporarily?).
- [X] Figure out if we want to keep the dodiff.sh after this or don't need it anymore and can delete it.
- [X] Implement some machinery (script?) that will combine template and diffs to produce the demo app, and figure out what our workflow will be there (and then document it also).
- [X] Figure out what we are doing with blog/ and e2e-tests: are we also doing blog via diffs, or just keeping it as it is? Are we deleting e2e-tests or also doing them via diffs?
  - Just keeping them for now as they are, no diffs, for start.
- [X] Update GitHub Actions to work with the new setup


.github/workflows/e2e-tests.yml
```diff
@@ -32,11 +32,13 @@ jobs:
...
-      # This step sets mock env vars in order to pass the validation steps so the app can run 
-      # in the CI environment. For env vars that are actually used in tests and therefore need real values, we set them in
-      # the GitHub secrets settings and access them in a step below.
+      # In order for the app to run in the dev mode we need to set the required env vars even if
+      # they aren't actually used by the app. This step sets mock env vars in order to pass the
+      # validation steps so the app can run in the CI environment. For env vars that are actually
+      # used in tests and therefore need real values, we set them in the GitHub secrets settings and
+      # access them in a step below.
       - name: Set required wasp app env vars to mock values
+        working-directory: ./template
```
Martinsos, line 41: Since I moved template to `template/`, I added `working-directory: ./template` to appropriate steps.

.github/workflows/retag-commit.yml
Martinsos, line 1: I deleted this one because this won't work for 0.13, so I think better to go manual for a bit and we can reconsider our approach later, maybe automate again.

CONTRIBUTING.md
Martinsos, line 1: Updated this to reflect new structure of the repo. Also made sure tehre is no duplicate info in top level README.

opensaas-sh/tools/dope.sh
Martinsos, line 1: Here is dope.sh, the main logic for diffing / patching.

opensaas-sh/tools/patch.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
Martinsos, line 1: Helper wrapper for patching.

opensaas-sh/tools/diff.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
Martinsos, line 1: Helper wrapper for dope.sh for diffing.

Martinsos (commented):
@vincanger take a look!



There is a lot o files in the PR that are just copies of blog or e2e-tests from deployed-version branch. SO those are not interesting, I didn't change anything there. Those are under blog/ or e2e-tests.

Also I just moved template files into `template/`.



Most interesting is stuff in opensaas-sh/app_diff, also some READMEs, opensaas-sh/tools, ... . I mostly commented on files that are interesting.

Notice that I also introduced sharing of .env.server file via dotenv-vault to make it easy for us to share env files, we will be doing this also in other Wasp projects, this is first application of it. You will see .env.vault file.

template/README.md
```diff
@@ -0,0 +1,8 @@
+# <YOUR_APP_NAME>
+
+This project is based on [OpenSaas](https://opensaas.sh) template and consists of three main dirs:
```
vincanger, line 3: The wording here is kind of confusing to me, after all, this _is_ the Open SaaS template, right?

What about changing it to:

"This is the [Open Saas](https://opensaas.sh) project directory, which consists of template itself (`app`), as well as two other dirs (`blog` and `e2e-tests1`" ?

vincanger: ok, I see. you're creating a template README for _their future app_.

opensaas-sh/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Development
+
+### Demo app (app_diff/)
+
+Since the demo app is just the open saas template with some small tweaks, and we want to be able to easily keep it up to date as the template changes, the demo app is defined via the diff against the template: `app_diff/`.
+
+Typical workflow:
+1. Run `./tools/patch.sh` to generate `app/` from `../template/` and `app_diff/`.
+2. If there are any conflicts (normally due to updates to the template), modify `app_diff/` till you resolve them.
```
sodic, line 17: How would we know whether there are any conflicts, who reports those and how? Is it a normal Git conflict or something else?

And how do we resolve them? By editing the diff files directly?

Also, what happens if the templates change in a way that makes the diff file outdated (e.g., the diff files says "remove this line", but the template no longer contains that line)?

opensaas-sh/README.md
vincanger, line 1: I think this workflow, and its explanation, are confusing to the uninitiated.

In step 1, I run `patch.sh` and generate the opensaas.sh app (aka demo app).
Then I have to generate a new `app_diff/` with `diff.sh`.

This is because we _do not_ commit an `app/` dir to version control. So maybe we should add this bit of info to the beginning of the workflow explanation. Something like:

"Since the demo app is just the open saas template with some small tweaks, and we want to be able to easily keep it up to date as the template changes, the demo app is defined via the diff against the template: `app_diff/`.

_Because we do not commit/push the generated `app/`

vincanger (approved):
Overall this good @Martinsos. I like the initiative here. The only thing is I fear we might be complicating things a bit, especially for outside contributors, just to get the demo app and the template in the same repo.

The other thing I fear is that, even though we direct people to use `wasp new -t saas` we will still get many users that clone the repo and then get all this extra stuff that confuses them.

But with that said, I think it's worth trying out, because it would make the workflow a lot easier from our side, granted that everything works (I also don't expect a lot of outside contributions to the demo app, to be honest).

I made a few small grammatical changes to READMEs, and a couple suggested changes, but I'm approving it here for now. Let's goooo 🐝

opensaas-sh/README.md
Martinsos: The `patch` reports them, and then you fix them in the diffs if I am correct (I keep forgetting, but I did it already. But will try it out much more in the following days).

Yeah, in that case the conflicts happens!

Truth be told, I am not 100% sure about the details of the workflow, I am also suspicious of how will this all go. I am quite sure conflicts will be happening as they should, but how complicated the resolution will be for us, and the whole workflow, is a bit hard to say at the moment. I would say there is 60% chance that this approach will be more trouble than it is worth. But let's give it a try, we have it working now, we will try to use it a bit, and if we don't like it, we can just generate the `app/` one final time, put that into separate repo and go with that. So I am looking at this as an experiment, at this point the best move is to give it a limited time try.

Even if experiment shows that this is not a good workflow for opensaas, it might still be useful for automating some of the e2e tests we will want to do for Wasp, so there is that also as a secondary use case.

template/README.md
Martinsos: Ah yes! So this is the README they get when they create the new app using the template: README for that app.

Martinsos:
> Overall this good @Martinsos. I like the initiative here. The only thing is I fear we might be complicating things a bit, especially for outside contributors, just to get the demo app and the template in the same repo.
> 
> The other thing I fear is that, even though we direct people to use `wasp new -t saas` we will still get many users that clone the repo and then get all this extra stuff that confuses them.
> 
> But with that said, I think it's worth trying out, because it would make the workflow a lot easier from our side, granted that everything works (I also don't expect a lot of outside contributions to the demo app, to be honest).
> 
> I made a few small grammatical changes to READMEs, and a couple suggested changes, but I'm approving it here for now. Let's goooo 🐝

Great! I did the changes you suggested above + checked the commits you did.

I agree this makes direct cloning even less viable, but I also think we shouldn't be trying to make cloning work, it makes the life for us too hard if we want to be able to maintain and test the template properly.

What is quite questionable is the whole diffing approach, but as you said let's give it a try and see: if it doesn't work for us I am more than happy to ditch it, it won't be hard to do.

opensaas-sh/README.md
sodic: Ok, if you're really stoked about trying it out, we can go for it.

That said, I'll voice my concerns just sin case you haven't thought about them.

I can't see any part of the procedure this approach has a chance of improving. 

> [!IMPORTANT]
> If, after reading what I think, you still see a potential benefit that I don't understand, you don't need to explain it to me (I'd be happy to read it, but it's not necessary if you don't have the time).
>
> I trust you to make a discretionary decision and go for it (assuming @vincanger is OK with it as well).

It sounds like this approach has all the problems/features of just keeping a separate directory plus:
 - The added inconvenience of dealing with patch and merge conflicts in diff files which are difficult for a human to read and write.
 - A worse experience during commits and code reviews - You have to compare diff files instead of proper source files
  - The lack of a human-readable example app in the repo - whoever wants to see the example app needs to check out the repo and construct it themselves. 
 - Extra technical/arcane steps when changing the example app - Instead of updating the example app directly, now you must:
   1. Construct the example app using a custom script and diff files.
   2. Update the generated example app changes (this is the only thing you had to do before).
   3. Construct the diffs from the updated example app.
   
   With updates that includes multiple iterations, this feels like a lot of pain: you update something and go through the procedure, you push it, the reviewer wants it changed, you update it again (and have to go through the procedure again), you push it again... 
   
   It's a similar problem as what we have with e2e tests, but worse. With e2e tests, you can avoid this cycle by waiting until the PR is finalized before generating the new tests. With this, you must do it every time. If you don't, the reviewer can't review your code. 

 
 All in all, since:
  - All the development is done in the generated app.
  - All the reviewing is done in the generated app.
  - All the testing is done in the generated app (don't know if we have tests yet, but that's where it's gonna happen).
  - All the reading (contributors, people wanting to use the template, etc.) is done in the generated app.

Martinsos: @sodic most of these are valid concerns! But, Chesterton's fence - if you don't know why it is there, there is probably a reason you are not seeing, right? So instead of saying we should keep a generated app, you should be asking what is it that you are not aware of, that made me go with this approach. Unless I love the complexity for the sake of it :D.

So, the reason why I did this whole diff thing instead of going with the generated appis that it is very hard to figure out how to update the generated app after you do updates to the template. I also wouldn't bother with diffs otherwise and would stick to normal code, of course.

Additionally, the part with reviewing the demo app, people looking at, ... -> those I don't think are concerns in practice, demo app is intended to be minimally different than template + for people to try it out, not to look at the code primarily.

We will see how this experiment goes, and as I mentioned previously I am aware of cons it introduces, but I don't see just having an app in there as a better solution. Instead, it might be better to just go with separate repo or something like that, although that has its own issues. I would consider having just the code in the dir next to it as the very last option, if nothing else works out.

PR #163, opened 2024-06-04
add cookie consent banner and tests

vincanger:
@martinsos because the project structure drifted so much, I decided to just create a new branch and manually add the changes from the original Cookie Consent PR https://github.com/wasp-lang/open-saas/pull/125


opensaas-sh/app_diff/.env.client.diff
```diff
@@ -1,7 +0,0 @@
---- template/app/.env.client
```
Martinsos: This even away, how is that? So demo app doesn't have .env.client version any more with this change. But I added it recently, because we need this stuff set up, right?

opensaas-sh/app_diff/migrations/migration_lock.toml.diff
Martinsos, line 1: Hm why did this file go away?

opensaas-sh/app_diff/package-lock.json.diff
Martinsos, line 1: So this one I am not quite sure if we want to version. Maybe we can. Not sure hm. If we didn't want to, what would we do? I guess gitignore it in the demo app. Ok, whatever we do is not terrible we can leave it and postpone decision till we figure it out.

opensaas-sh/app_diff/main.wasp.diff
```diff
@@ -87,10 +85,17 @@
...
+   createdAt                 DateTime        @default(now())
+ 
+-  email                     String?         @unique
+-  username                  String?         @unique
++  email                     String?         
++  username                  String?         
    lastActiveTimestamp       DateTime        @default(now())
-   isAdmin                   Boolean         @default(false)
+-  isAdmin                   Boolean         @default(false)
++  isAdmin                   Boolean         @default(true)
```
Martinsos, line 98: These changes are on purpose? You removed @unique from email and username, and also made isAdmin be true by default, in this PR? Doesn't sound very connected with the cookie banner though hm. Maybe you were fixing something I missed when doing initial creation of app_diff?

opensaas-sh/app_diff/src/client/components/cookie-consent/Config.ts.diff
```diff
@@ -0,0 +1,11 @@
+--- template/app/src/client/components/cookie-consent/Config.ts
++++ opensaas-sh/app/src/client/components/cookie-consent/Config.ts
+@@ -94,7 +94,7 @@
+               'We use cookies primarily for analytics to enhance your experience. By accepting, you agree to our use of these cookies. You can manage your preferences or learn more about our cookie policy.',
+             acceptAllBtn: 'Accept all',
+             acceptNecessaryBtn: 'Reject all',
+-            // showPreferencesBtn: 'Manage Individual preferences', // (OPTIONAL) Activates the preferences modal
++            // showPreferencesBtn: 'Manage Individual preferences',
```
Martinsos, line 8: So in demo app you just removed the comment on the right. I would put that comment back, so we don't have this diff file, to keep it simple. It is demo app, it is ok if comments are left. We are striving for minimal changes.

opensaas-sh/app_diff/src/server/auth/setUsername.ts.diff
```diff
@@ -0,0 +1,26 @@
+--- template/app/src/server/auth/setUsername.ts
++++ opensaas-sh/app/src/server/auth/setUsername.ts
+@@ -4,7 +4,6 @@
+ 
+ export const getEmailUserFields = defineUserSignupFields({
+   username: (data: any) => data.email,
+-  isAdmin: (data: any) => adminEmails.includes(data.email),
```
Martinsos, line 7: Is this also something I forgot and you are fixing now? Because it also doesn't look connected to the cookie banner.

opensaas-sh/blog/src/content/docs/guides/analytics.md
```diff
@@ -7,9 +7,9 @@ banner:
 ---
 This guide will show you how to integrate analytics for your app. You can choose between [Google Analytics](#google-analytics) and [Plausible](#plausible).
 
-Google Analytics is free, but tends to be more cumbersome to use.
+Google Analytics is free, but uses cookies, so you must implement the [Cookie Consent Modal](./cookie-consent.md) when using it.
```
Martinsos, line 10: From what I read, in EU this is a must, but outside it depends on state / country ... . So maybe we could, instead of `must`, go with something like `you will probably need/want to`.

opensaas-sh/app_diff/.env.client.diff
vincanger, line 1: hm. it did this because on the `template` I have an `.env.client` file set with same env vars as for the demo app, bu this file is under the .gitingore list and doesn't get pushed to git it, which makes sense for the `template` dir. But we want the `.env.client` file to be pushed in the `opensaas-sh` dir.

opensaas-sh/blog/src/content/docs/guides/analytics.md
```diff
@@ -7,9 +7,9 @@ banner:
...
 
-Plausible is an open-source, privacy-friendly alternative to Google Analytics. It's also easier to use than Google if you use their hosted service, which is a paid feature. But, it is completely free if you want to self-host it, although this comes with some additional setup steps.
+Plausible is an open-source, privacy-friendly alternative to Google Analytics. **You DO NOT have to use the cookie consent modal** with Plausible, as it does not use cookies. It's also easier to use than Google if you use their hosted service, but be aware it is a paid feature. It is completely free if you want to self-host it, although this comes with some additional setup steps.
```
Martinsos, line 12: Ok nice explanation!

opensaas-sh/blog/src/content/docs/guides/cookie-consent.md
```diff
@@ -0,0 +1,159 @@
...
+
+```sh
+_ga
+_ga... # Google Analytics cookies.
+cc_cookie # Cookie Consent cookie. The name of this cookie can be changed in the config file.
+```
+
+### Plausible Analytics
+
+If you decide to go with [Plausible Analytics](./analytics.md/#plausible), you **DO NOT** need to ask users for their consent to use cookies becuase Plausible, as a privacy-first analytics provider, [does not use cookies](https://plausible.io/privacy-focused-web-analytics). Instead, It collects website usage data anonymously and in aggregate form only, without any personally identifiable information
```
Martinsos, line 49:
```suggestion
If you decide to go with [Plausible Analytics](./analytics.md/#plausible), you **DO NOT** need to ask users for their consent to use cookies because Plausible, as a privacy-first analytics provider, [does not use cookies](https://plausible.io/privacy-focused-web-analytics). Instead, It collects website usage data anonymously and in aggregate form only, without any personally identifiable information
```

opensaas-sh/blog/src/content/docs/guides/cookie-consent.md
Martinsos, line 1: Wow ok, extensive docs, impressive! We will need to maintain this :D. I am always tempted to write a bit less and link a bit more, but, there is value to having more stuff in the docs for sure. I am writing this just as a side thought.

template/app/main.wasp
```diff
@@ -15,12 +15,12 @@ app OpenSaaS {
         "<meta name='twitter:image:width' content='800' />",
         "<meta name='twitter:image:height' content='400' />",
         "<meta name='twitter:card' content='summary_large_image' />",
-        // TODO: You can put your analytics scripts below (https://docs.opensaas.sh/guides/analytics/):
-        //   If you are going with Plausible:
+        // TODO: You can put your Plausible analytics scripts below (https://docs.opensaas.sh/guides/analytics/):
```
Martinsos, line 18: Aha, so we are not recommending them to put google analytics script here anymore. Ok, that makes sense!

template/app/package-lock.json
Martinsos, line 1: Ok yeah so this one also I am not sure if we want to version it or not. Hmmmm. If not, we would ignore it on the higher level, same way we ignore migrations, not here. I would probably go for ignoring it. You know what forget it for now, I will open a new issue for this.

template/blog/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "docs",
+  "name": "blog2",
```
Martinsos, line 2: Hah why `2`?

Martinsos (approved):
Approving since only small stuff is left, nice!

opensaas-sh/app_diff/.env.client.diff
Martinsos: Hm interesting. But dope.sh takes that into account, it looks only at notgitignored files. So if you have it gitignored in tempalte/app, and it is not ignored in opensaas-sh, then it shouldn't remove it, instead it should see that file missing on the template side and existing on the opensaas-sh side. So either that is working wrong (although I don't think so), or there is some other cause.

vincanger: I only got the diff to fix by removing the .env.client file from the template :( so it seems like it's not working

opensaas-sh/app_diff/main.wasp.diff
vincanger: Yeah. We want every user to be an Admin on the demo app so they can test out the Admin dashboard :)

opensaas-sh/blog/src/content/docs/guides/cookie-consent.md
vincanger: It's easy to delete stuff :) we can leave it for now but if gets annoying, we can remove it

Martinsos: Makes sense, great work!

opensaas-sh/app_diff/.env.client.diff
Martinsos: Ok we found bug in dope.sh, will be fixing it! Thanks @vincanger !

PR #167, opened 2024-06-06
Fixed bug in dope.sh with not generating diffs for gitignored base files.

Martinsos:
Fixed the bug + added some general comments to make code easier to understand (it is bash, at the end).


vincanger (approved):

vincanger:
nice. gave it a try and it worked :)

PR #170, opened 2024-06-07
Grouped all file-upload functionality.

Martinsos:
I started doing some vertical reorganization @vincanger !

I thought first to accumulate it all in one big PR, but that will be huge, so I thought maybe better to do it in chunks like this? Anyway we are not releasing automatically, so we are ok with open-saas being half-converted for some time on `main`, till we vertically organize it whole.

Take a look at this and tell me how you like it. This is file upload stuff vertically organized. I quite like it. If you also like it, I will continue doing this with the rest of the template.

TODO:
- [ ] Update demo app if needed
- [ ] Update docs if needed


template/app/main.wasp
```diff
@@ -444,3 +411,40 @@ job dailyStatsJob {
   },
   entities: [User, DailyStats, Logs, PageViewSource]
 }
+
+
+//#region File Upload
```
Martinsos, line 416: Here I grouped all file upload stuff.

template/app/src/file-upload/FileUploadPage.tsx
Martinsos, line 1: This file got moved to src/file-upload .

template/app/src/file-upload/operations.ts
Martinsos, line 1: This is a new file where I put both queries and actions for file upload.

template/app/src/file-upload/s3Utils.ts
Martinsos, line 1: This file just got moved, to src/file-upload.

template/app/src/server/actions.ts
Martinsos, line 1: I moved file upload action(s) from this file to src/file-upload/operations.ts.

template/app/src/server/queries.ts
Martinsos, line 1: I moved file upload queries from this file to src/file-upload/operations.ts.

template/app/main.wasp
vincanger: cool. I like it.

vincanger (changes requested):
Noice. I like it @Martinsos. This will make a lot of sense when multiple `wasp` files becomes a feature.

We should probably do these two things:
1. create some file upload e2e-tests 
2. update the docs with the new file structure.

I can take on #1 if you like (and add it into this PR -- that's possible if i just checkout this PR and make commits to it, right? not sure if that's a good practice though).

And we can do  #2 at the end when the entire structure is grouped by feature. What do you think? I'll create an issue for that now.

Martinsos:
@vincanger awesome!

So e2e tests -> if we had none so far, then let's leave that for a separate issue, to add e2e tests.

As for docs, since we had them already, then let's do that change as the part of this PR. If you want, you can add them by adding commits to this PR, but I can also do it.

opensaas-sh/blog/src/content/docs/guides/file-uploading.md
```diff
@@ -119,26 +119,13 @@ With your S3 bucket set up and your AWS credentials in place, you can now start
 
 To begin customizing file uploads, is important to know where everything lives in your app. Here's a quick overview:
 - `main.wasp`:
-  - The `File entity` can be found here. Here you can modify the fields to suit your needs:
```
Martinsos, line 122: I removed the code snippet for File entity here, it is not needed, they can check it in main.wasp file, and we just need to maintain it.

Martinsos:
@vincanger I updated docs and I also updated the demo app! Btw I noticed that demo app doesn't patch correctly, some changes were pushed to main without updating the demo app (app_diff), but we should fix that in some other PR.

vincanger (approved):
Nice! Looks good. I took care of the app_diff update here: https://github.com/wasp-lang/open-saas/pull/207

Issue #174, opened 2024-06-10
e2e tests failing with cookie banner check


Martinsos:
![image](https://github.com/wasp-lang/open-saas/assets/1536647/f8a03136-90cb-45b4-8935-fb8bfc42bf65)

E2e tests often fail on cookie banner test since we added it. It seems to be flaky because it can help if we restart the github action.

Btw that assertion that fails, I would also rewrite it to be `expect(gaCookiesArr.length).toBe(2)` (made it up a bit but I am sure something like this is doable), so we get more info when the test fails.

vincanger:
fixed in https://github.com/wasp-lang/open-saas/pull/176

PR #176, opened 2024-06-13
avoid cookie setting race condition

vincanger:
## Description

tests were failing for cookie banner check https://github.com/wasp-lang/open-saas/issues/174

## Contributor Checklist

### DIFF
--- a/template/e2e-tests/tests/landingPageTests.spec.ts
+++ b/template/e2e-tests/tests/landingPageTests.spec.ts
@@ -57,16 +57,17 @@ test.describe('cookie consent tests', () => {
     expect(cookieObject.categories.includes('analytics')).toBeTruthy();
 
     // Wait for Google Analytics cookies to be set after accepting
-    const MAX_TIME = 45000;
+    const MAX_TIME_MS = 45000;
     const startTime = Date.now();
     while (cookies.length === 1) {
-      if (Date.now() - startTime > MAX_TIME) {
+      if (Date.now() - startTime > MAX_TIME_MS) {
         throw new Error('Timeout: Google Analytics cookies not set.');
       }
       cookies = await context.cookies();
+      await new Promise((resolve) => setTimeout(resolve, 1000));
     }
     let gaCookiesArr = await context.cookies(); // Call the cookies method again to avoid race condition
     gaCookiesArr = gaCookiesArr.filter((c) => c.name.startsWith('_ga'));
-    expect(gaCookiesArr.length === 2).toBeTruthy();
+    expect(gaCookiesArr.length).toBe(2)
   });
 });
--- a/template/e2e-tests/tests/landingPageTests.spec.ts
+++ b/template/e2e-tests/tests/landingPageTests.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from '@playwright/test';
+import { test, expect, Cookie } from '@playwright/test';
 
 const DOCS_URL = 'https://docs.opensaas.sh';
 
@@ -57,17 +57,25 @@ test.describe('cookie consent tests', () => {
     // Check that the Cookie Consent cookie is set. This should happen immediately, and then the GA cookies will get set after it, dynamically.
     expect(cookieObject.categories.includes('analytics')).toBeTruthy();
 
-    // Wait for Google Analytics cookies to be set after accepting
-    const MAX_TIME_MS = 45000;
+    const areGaCookiesSet = (cookies: Cookie[]) => {
+      const gaCookiesArr = cookies.filter((c) => c.name.startsWith('_ga'));
+      return gaCookiesArr.length === 2; // GA cookies are _ga and _ga_<GA_ANALYTICS_ID>
+    };
+
     const startTime = Date.now();
-    while (cookies.length < 3) { // 3 cookies will be set: cc_cookie, _ga, _ga_xxx
-      if (Date.now() - startTime > MAX_TIME_MS) {
-        throw new Error('Timeout: Google Analytics cookies not set.');
-      }
-      await new Promise((resolve) => setTimeout(resolve, 1000));
+    const MAX_TIME_MS = 45000;
+    let timeElapsed = 0;
+
+    while (!areGaCookiesSet(cookies) && timeElapsed < MAX_TIME_MS) {
+      await new Promise((resolve) => setTimeout(resolve, 1000)); // wait for 1 second before checking again
+      timeElapsed = Date.now() - startTime;
       cookies = await context.cookies();
     }
-    let gaCookiesArr = cookies.filter((c) => c.name.startsWith('_ga'));
-    expect(gaCookiesArr.length).toBe(2);
+
+    if (timeElapsed >= MAX_TIME_MS) {
+      throw new Error('Timeout: Google Analytics cookies not set.');
+    }
+
+    expect(areGaCookiesSet(cookies)).toBeTruthy();
   });
 });
--- a/template/e2e-tests/tests/landingPageTests.spec.ts
+++ b/template/e2e-tests/tests/landingPageTests.spec.ts
@@ -60,15 +60,14 @@ test.describe('cookie consent tests', () => {
     // Wait for Google Analytics cookies to be set after accepting
     const MAX_TIME_MS = 45000;
     const startTime = Date.now();
-    while (cookies.length === 1) {
+    while (cookies.length < 3) { // 3 cookies will be set: cc_cookie, _ga, _ga_xxx
       if (Date.now() - startTime > MAX_TIME_MS) {
         throw new Error('Timeout: Google Analytics cookies not set.');
       }
-      cookies = await context.cookies();
       await new Promise((resolve) => setTimeout(resolve, 1000));
+      cookies = await context.cookies();
     }
-    let gaCookiesArr = await context.cookies(); // Call the cookies method again to avoid race condition
-    gaCookiesArr = gaCookiesArr.filter((c) => c.name.startsWith('_ga'));
+    let gaCookiesArr = cookies.filter((c) => c.name.startsWith('_ga'));
     expect(gaCookiesArr.length).toBe(2);
   });
 });
--- a/template/e2e-tests/tests/landingPageTests.spec.ts
+++ b/template/e2e-tests/tests/landingPageTests.spec.ts
@@ -64,10 +64,9 @@ test.describe('cookie consent tests', () => {
         throw new Error('Timeout: Google Analytics cookies not set.');
       }
       cookies = await context.cookies();
-      console.log('cookies >>> ', cookies)
     }
-    
-    const gaCookiesArr = cookies.filter((c) => c.name.startsWith('_ga'));
+    let gaCookiesArr = await context.cookies(); // Call the cookies method again to avoid race condition
+    gaCookiesArr = gaCookiesArr.filter((c) => c.name.startsWith('_ga'));
     expect(gaCookiesArr.length === 2).toBeTruthy();
   });
 });
--- a/template/e2e-tests/tests/landingPageTests.spec.ts
+++ b/template/e2e-tests/tests/landingPageTests.spec.ts
@@ -63,19 +63,15 @@ test.describe('cookie consent tests', () => {
     };
 
     const startTime = Date.now();
-    const MAX_TIME_MS = 45000;
+    const MAX_TIME_MS = 10000;
     let timeElapsed = 0;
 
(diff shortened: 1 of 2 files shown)


template/e2e-tests/tests/landingPageTests.spec.ts
```diff
@@ -57,16 +57,16 @@ test.describe('cookie consent tests', () => {
     expect(cookieObject.categories.includes('analytics')).toBeTruthy();
 
     // Wait for Google Analytics cookies to be set after accepting
-    const MAX_TIME = 15000;
+    const MAX_TIME = 45000;
```
Martinsos, line 60: Is this milliseconds? If so, I would name it `MAX_TIME_MS`. Or, I would go with `MAX_TIME_SECS=45` and then down there, do `Date.now() - startTime > MAX_TIME_SECS * 1000`.

Btw how did you come up with 45 seconds? Empirically :D?

template/e2e-tests/tests/landingPageTests.spec.ts
```diff
@@ -57,16 +57,16 @@ test.describe('cookie consent tests', () => {
     expect(cookieObject.categories.includes('analytics')).toBeTruthy();
```
Martinsos, line 57: Aha so this cc_cookie, that one will be available immediatell? What is this cookie, and how it is that this one will be availble right now but ga cookie won't?

template/e2e-tests/tests/landingPageTests.spec.ts
```diff
@@ -57,16 +57,16 @@ test.describe('cookie consent tests', () => {
...
     // Wait for Google Analytics cookies to be set after accepting
-    const MAX_TIME = 15000;
+    const MAX_TIME = 45000;
     const startTime = Date.now();
     while (cookies.length === 1) {
       if (Date.now() - startTime > MAX_TIME) {
         throw new Error('Timeout: Google Analytics cookies not set.');
       }
       cookies = await context.cookies();
     }
```
Martinsos, line 67: So you are consuming 100% CPU time here: doing check by check, as soon as you can.
I am not sure in this case, in this specific test, how bad is that, is there anything else that would ideally be executing next to this and is being blocked out by us hogging the CPU here, but I would throw in a `await new Promise(resolve => setTimeout(resolve, 1000))`.
The only reason why this thing is not currently completely hogging the CPU is this `cookies = await context.cookies()` -> this pops this current execution from the execution stack or whatever is the name and allows some other function to get opportunity to run. If you didn't have this, nothing else could even try to run for 45 second potentially.

template/e2e-tests/tests/landingPageTests.spec.ts
```diff
@@ -57,16 +57,16 @@ test.describe('cookie consent tests', () => {
...
-
-    const gaCookiesArr = cookies.filter((c) => c.name.startsWith('_ga'));
+    let gaCookiesArr = await context.cookies(); // Call the cookies method again to avoid race condition
```
Martinsos, line 68: What does this mean, which race condition do you avoid this way?

template/e2e-tests/tests/landingPageTests.spec.ts
```diff
@@ -57,16 +57,16 @@ test.describe('cookie consent tests', () => {
...
+    gaCookiesArr = gaCookiesArr.filter((c) => c.name.startsWith('_ga'));
     expect(gaCookiesArr.length === 2).toBeTruthy();
```
Martinsos, line 70: Change this to

```suggestion
    expect(gaCookiesArr.length).toBe(2);
```

in order to get more info if test fails.

template/e2e-tests/tests/landingPageTests.spec.ts
vincanger: cc_cookie is the one set by the Cookie Consent banner component. It contains the settings for which cookies should be set or not. I guess this cookie gets read first, and then sets the google analytics script dynamically, which is why it takes a before the GA cookies get set.

template/e2e-tests/tests/landingPageTests.spec.ts
vincanger: this is probably why I was getting a race condition: the while loop condition was being met, but then the `cookies` array being checked by `expect` only contained one cookie, the `cc_cookie`

template/e2e-tests/tests/landingPageTests.spec.ts
vincanger: i thought at first it was failing due to a timeout, so I just increased the time based on vibez

template/e2e-tests/tests/landingPageTests.spec.ts
Martinsos: Ok makes sense! Maybe you could write a comment about it here, becuase whoever reads this, will have hard time figuring that out on their own. SOmethig short: `Check that Cookie Consent cookie is set. This should happen immediately, and then the GA cookie will get set after it, dynamically`.

template/e2e-tests/tests/landingPageTests.spec.ts
Martinsos: Hm the race condition still doesn't make any sense, it can't have 2 cookies at one moment and then have 1 cookie again. Why do you think you were having race condition, how was that manifesting exactly?

vincanger: I think it has to do with how Playwright, and it's browserContext, fetch the cookies, but I'm not sure. If I remove the second `await context.cookies()`, I get the following error:

```ts
Error: expect(received).toBe(expected) // Object.is equality

Expected: 2
Received: 1
```

and the only cookie inside the `cookies` array is the `cc_cookie`. 

When I refetch the cookies with the second await (line 71), it passes.

Martinsos: Hm that doesn't seem possible though.

Execution can continue after the while loop only if `cookies` had not 1 element. So hopefully 2 and not 0 or 3 or something :). That is how our logic is written. If time goes out, error is thrown, so no execution will happen after that.

So, if `cookie` var has two elements, and the next line is using `cookie` var directly, there is no way that something changed it in between. In theory, yeah, `cookie` is just pointing to an array with two elements and sometihng else could have manipulated that array. But JS is single threaded, you don't have any await or anything in between the while loop condition passing and then looking into `cookie` to get ga cookie, so I don't see how it would theoretically be possible that this happens. Regardless of whatever Playwright might be doing, there are rules to JS that it can't go around.

Are you sure this is still happenign? Can you replicate and send me the exact code that you managed to replicate this with?

Martinsos: Actually now it is possible in theory since you put await setTimeout after getting the cookies. Although I would find that very surprising, but theoretically possible.
Put `await setTimeout` before `await cookies`, then try to replicate as I suggested above, it shouldn't theoretically be able to happen.

Martinsos: Ok wait actually @vincanger , I just realized you expect the filtered cookies to be of length 2.

OK, so check this out:
1. You first get that cc cookie. 
2. Then you wait for it to fetch some more cookies in that while loop, so you have more than one cookie.
3. Finally, you filter out ga cookies, and then you check that you have 2 of ga cookies (ga, not all cookies!).

So this is your logic. Is this really want you want though? So there should be 3 cookies all together, 1 cc cookie and 2 ga cookies? Or is there a mistake here?

vincanger: > Ok wait actually @vincanger , I just realized you expect the filtered cookies to be of length 2.
> 
> OK, so check this out:
> 
> 1. You first get that cc cookie.
> 2. Then you wait for it to fetch some more cookies in that while loop, so you have more than one cookie.
> 3. Finally, you filter out ga cookies, and then you check that you have 2 of ga cookies (ga, not all cookies!).

yeah this was my mistake.

template/e2e-tests/tests/landingPageTests.spec.ts
```diff
@@ -54,19 +54,21 @@ test.describe('cookie consent tests', () => {
...
     const consentCookie = cookies.find((c) => c.name === 'cc_cookie');
     const cookieObject = JSON.parse(decodeURIComponent(consentCookie.value));
+    // Check that the Cookie Consent cookie is set. This should happen immediately, and then the GA cookies will get set after it, dynamically.
     expect(cookieObject.categories.includes('analytics')).toBeTruthy();
 
     // Wait for Google Analytics cookies to be set after accepting
-    const MAX_TIME = 15000;
+    const MAX_TIME_MS = 45000;
     const startTime = Date.now();
     while (cookies.length === 1) {
```
Martinsos: Btw this whole thing with checking length === 1 and assuming that will tell us if ga cookies are set is a bit flimsy. What if some other cookie was set by some other piece of Wasp, right at this moment? Not very likely, but could happen. What if cookies get set to 0 for a short while? Shouldn't happen but still it is weird.

What you really want, is to check if ga cookie is set. So if you want to check that, let's check that! Write an anonymous function called `isGaCookieSet(cookies)` and use that in the while loop as a condition. Once that checks out or time runs out, stop the while loop. You can check do `expect` where you just check if time spent was too long or not.

template/e2e-tests/tests/landingPageTests.spec.ts
Martinsos: Took me quite a while to figure this out also, it is not easy to notice! This is really the consequence of not-clean-code -> playing with lengths and making assumptions around that. It made the code complex enough for us both to not notice this mistake for some time. Simpler code -> less place for mistake. Which is now why I made this comment: https://github.com/wasp-lang/open-saas/pull/176#discussion_r1638129621 -> let's do it clean code style.

vincanger: functional programming ftw

template/e2e-tests/tests/landingPageTests.spec.ts
```diff
@@ -57,17 +57,25 @@ test.describe('cookie consent tests', () => {
...
       cookies = await context.cookies();
     }
-    let gaCookiesArr = cookies.filter((c) => c.name.startsWith('_ga'));
-    expect(gaCookiesArr.length).toBe(2);
+
+    if (timeElapsed >= MAX_TIME_MS) {
+      throw new Error('Timeout: Google Analytics cookies not set.');
+    }
+
+    expect(areGaCookiesSet(cookies)).toBeTruthy();
```
Martinsos, line 79: The way you have it, this expect will always be truth, so it is not really doing anything, except for telling playwright that test passed.

You could do this though:

```suggestion
    expect(timeElapsed).toBeLessThan(MAX_TIME_MS);
```

template/e2e-tests/tests/landingPageTests.spec.ts
```diff
@@ -57,17 +57,25 @@ test.describe('cookie consent tests', () => {
...
+    };
+
     const startTime = Date.now();
-    while (cookies.length < 3) { // 3 cookies will be set: cc_cookie, _ga, _ga_xxx
-      if (Date.now() - startTime > MAX_TIME_MS) {
-        throw new Error('Timeout: Google Analytics cookies not set.');
-      }
-      await new Promise((resolve) => setTimeout(resolve, 1000));
+    const MAX_TIME_MS = 45000;
+    let timeElapsed = 0;
```
Martinsos, line 67: How about we rename this one to `timeElapsedTillGaCookiesSet`

template/e2e-tests/tests/landingPageTests.spec.ts
```diff
@@ -57,17 +57,25 @@ test.describe('cookie consent tests', () => {
...
+      return gaCookiesArr.length === 2; // GA cookies are _ga and _ga_<GA_ANALYTICS_ID>
+    };
+
     const startTime = Date.now();
-    while (cookies.length < 3) { // 3 cookies will be set: cc_cookie, _ga, _ga_xxx
-      if (Date.now() - startTime > MAX_TIME_MS) {
-        throw new Error('Timeout: Google Analytics cookies not set.');
-      }
-      await new Promise((resolve) => setTimeout(resolve, 1000));
+    const MAX_TIME_MS = 45000;
```
Martinsos, line 66: Can we now take this one down to something like `10000`? Because this wasn't the cause at the end, right? 10 seconds sounds like a ton of time to set some cookies.

vincanger:
ok. nice suggestions @Martinsos

template/e2e-tests/tests/landingPageTests.spec.ts
Martinsos: @vincanger ha indeed!

Martinsos (approved):
Nice, approved!

Issue #180, opened 2024-06-13
onAuthSucceededRedirectTo value isn't respected when the user visits login route post successful auth


Case-E:
**Describe the bug**
When the user first logs in (tested email auth), the user is redirected to the value set in `onAuthSucceededRedirectTo ` or the default `/` if nothing is set. However, once successfully logged in, if the user visits the login route again, they are redirected to `/` instead of the path set in `onAuthSucceededRedirectTo `. 

**To Reproduce**
1. Set the `onAuthSucceededRedirectTo` path in the auth method. 
2. Go to the login route Ex: `/login`
3. Login to a user account (using email auth)
4. Navigate to the `/login` route again

**Expected behavior**
User should be navigated to the path set in `onAuthSucceededRedirectTo`

**Desktop (please complete the following information):**
 - OS: macOS Sonoma 14.5 (23F79)
 - Browser Safari
 - Version Version 17.5 (19618.2.12.11.6)

Martinsos:
@vincanger what do you think?

vincanger:
yep. I'll remove that now @Martinsos. Thanks for catching this @Case-E.

If you still wanna redirect logged in users, the best way is actually to do this:

```tsx
import { Redirect } from 'react-router-dom';

export default function Login() {

  const { data: user } = useAuth();

  if (user) {
    return <Redirect to='/demo-app' />;
  }
```

vincanger:
here is the PR for removing the redirect and creating a new diff for opensaas.sh https://github.com/wasp-lang/open-saas/pull/182

Case-E:
Thanks @Martinsos and @vincanger! :D

Martinsos:
Thank you @Case-E !

PR #182, opened 2024-06-20
Remove redirecting logged in user on Login page

vincanger:
https://github.com/wasp-lang/open-saas/issues/180


Martinsos (approved):
Perfect!

PR #184, opened 2024-06-20
add instructions for local e2e env vars

vincanger:
instructed users to add `SKIP_EMAIL_VERIFICATION_IN_DEV=true` to `.env.server` before running local e2e tests to automatically verify the test user on signup


Martinsos:
I commented on the origianl issue, I think there might potentially be a better solution.

Martinsos (approved):

PR #191, opened 2024-06-24
Add note on `sudo` for Stripe CLI 

vincanger:
A user on WSL encountered errors running the Stripe CLI: https://discord.com/channels/686873244791210014/1253545019785351180/1253951893877817527

The workaround was to append `sudo` to the commands as referenced in the GitHub issue (provided in the docs)

### DIFF
--- a/opensaas-sh/blog/src/content/docs/guides/stripe-integration.md
+++ b/opensaas-sh/blog/src/content/docs/guides/stripe-integration.md
@@ -95,6 +95,21 @@ or for other install scripts or OSes, follow the instructions [here](https://str
 
 Now, let's start the webhook server and get our webhook signing secret.
 
+First, login:
+```sh
+stripe login
+```
+
+:::caution[Errors running the Stripe CLI]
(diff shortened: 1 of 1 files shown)


opensaas-sh/blog/src/content/docs/guides/stripe-integration.md
```diff
@@ -95,6 +95,21 @@ or for other install scripts or OSes, follow the instructions [here](https://str
 
 Now, let's start the webhook server and get our webhook signing secret.
 
+First, login:
+```sh
+stripe login
+```
+
+:::caution[Errors running the Stripe CLI]
+If you're seeing errors, try appending `sudo` to the stripe commands:
```
Martinsos, line 104:
```suggestion
If you're seeing errors, consider appending `sudo` to the stripe commands:
```

A bit of nitpicking, but running sometihng with `sudo` is always a bit scary so this emphasies they should be think carefully about this and not just give it a try.

Martinsos (approved):
@vincanger approved but also left two comments, resolve as you see fit!

Issue #160, opened 2024-05-31
setUsername.ts does more than the name says


Martinsos:
Before it was doing just setting username, but now it also sets fields for social auth.

We should either update the name of the file to capture that also, or if that is not viable, split the logic into multiple files.

PR #193, opened 2024-06-26
change filenames for auth user fields 

vincanger:
issue https://github.com/wasp-lang/open-saas/issues/160


template/app/main.wasp
```diff
@@ -36,23 +36,23 @@ app OpenSaaS {
         },
         emailVerification: {
           clientRoute: EmailVerificationRoute,
-          getEmailContentFn: import { getVerificationEmailContent } from "@src/server/auth/email.js",
+          getEmailContentFn: import { getVerificationEmailContent } from "@src/server/auth/emailContentFns",
```
Martinsos, line 39: I think email.js was actually a better name. A bit hard to exactly say why, but `emailContentFn` feels quite overly specific, to the degree that I am looking for meaning in that name that is just not there.
`auth/email` is fine -> obviusly some stuff for email auth is there. Both fo these functions that are about email contents are small enough and too specific to not deserve their own file currently. So I would revert this to email.js.

template/app/main.wasp
```diff
@@ -36,23 +36,23 @@ app OpenSaaS {
...
         },
         passwordReset: {
           clientRoute: PasswordResetRoute,
-          getEmailContentFn: import { getPasswordResetEmailContent } from "@src/server/auth/email.js",
+          getEmailContentFn: import { getPasswordResetEmailContent } from "@src/server/auth/emailContentFns",
         },
-        userSignupFields: import { getEmailUserFields } from "@src/server/auth/setUsername.js",
+        userSignupFields: import { getEmailUserFields } from "@src/server/auth/setUserSignupFields",
```
Martinsos, line 45: setUserSignupFields -> why `set`? We had `setUsername` before, but that name wasn't good. And looking into file, there is no `set`, it is all `get` instead! So better would be `getUserSignupFields` I think?

But we can go a bit less specific and just go with `userSignupFields` as a filename, that should be fine I think.

My problem here is actually that `userSignupFIelds` in general is a bit weird term / naming, feels like we need  a better name for it. But it wasn't so easy to find a better name. I expect this part of Wasp to evolve with time into a better abastraction that will naturally get a better name then, or just that we find a better name. But in the meantime, yeah I think `userSignupFields` is ok.

Martinsos (changes requested):

Martinsos (approved):

Issue #146, opened 2024-05-23
Clean up usersSeed.ts


Martinsos:
Clean up the code a bit. Nicer comment, in a better place. Consider if we can put main function on top maybe, to read top down like a story.

PR #197, opened 2024-06-27
clean up db seeding file

vincanger:
https://github.com/wasp-lang/open-saas/issues/146

### DIFF
--- a/template/app/main.wasp
+++ b/template/app/main.wasp
@@ -64,7 +64,7 @@ app OpenSaaS {
     // Run `wasp db seed` to seed the database with the seed functions below:
     seeds: [
       // Populates the database with a bunch of fake users to work with during development.
-      import { devSeedUsers } from "@src/server/scripts/usersSeed.js",
+      import { seedMockUsersToDB } from "@src/server/scripts/dbSeeds",
     ]
   },
 
--- a/template/app/main.wasp
+++ b/template/app/main.wasp
@@ -100,8 +100,8 @@ entity User {=psl
 
   stripeId                  String?
   checkoutSessionId         String?
-  subscriptionTier          String?
-  subscriptionStatus        String?
+  subscriptionStatus        String?         // 'active', 'canceled', 'past_due', 'deleted', null
+  subscriptionTier          String?         // 'hobby-tier', 'pro-tier', null
   sendEmail                 Boolean         @default(false)
   datePaid                  DateTime?
   credits                   Int             @default(3)
--- a/template/app/main.wasp
+++ b/template/app/main.wasp
@@ -64,7 +64,7 @@ app OpenSaaS {
     // Run `wasp db seed` to seed the database with the seed functions below:
     seeds: [
       // Populates the database with a bunch of fake users to work with during development.
-      import { seedMockUsersToDB } from "@src/server/scripts/dbSeeds",
+      import { seedMockUsers } from "@src/server/scripts/dbSeeds",
     ]
   },
 
--- a/template/app/src/server/scripts/dbSeeds.ts
+++ b/template/app/src/server/scripts/dbSeeds.ts
@@ -9,7 +9,9 @@ import { TierIds } from '../../shared/constants.js';
  * For more info see: https://wasp-lang.dev/docs/data-model/backends#seeding-the-database
  */
 export async function seedMockUsers(prismaClient: PrismaClient) {
-  await Promise.all(generateMockUsersData(50).map((user) => prismaClient.user.create({ data: user })));
+  await Promise.all(generateMockUsersData(50).map(
+    data => prismaClient.user.create({ data })
+  ));
 }
 
 function generateMockUsersData(numOfUsers: number) {
--- a/template/app/src/server/scripts/dbSeeds.ts
+++ b/template/app/src/server/scripts/dbSeeds.ts
@@ -4,53 +4,38 @@ import type { PrismaClient } from '@prisma/client';
 import { TierIds } from '../../shared/constants.js';
 
 /**
- * This function, which we've imported in `app.db.seeds` in the `main.wasp` file, 
+ * This function, which we've imported in `app.db.seeds` in the `main.wasp` file,
  * seeds the database with mock users via the `wasp db seed` command.
  * For more info see: https://wasp-lang.dev/docs/data-model/backends#seeding-the-database
  */
-export async function seedMockUsersToDB(prismaClient: PrismaClient) {
-  const mockUsers: Omit<User, 'id'>[] = getMockUsers(50);
-  const persistUsersToDB = mockUsers.map((user) => {
-    return prismaClient.user.create({
-      data: user,
-    });
-  });
-
-  try {
-    await Promise.all(persistUsersToDB);
-  } catch (error) {
-    console.error(error);
-  }
+export async function seedMockUsers(prismaClient: PrismaClient) {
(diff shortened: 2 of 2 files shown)


7AXEL:
Check out this project: https://github.com/7AXEL/Web2APK

template/app/src/server/scripts/dbSeeds.ts
```diff
@@ -3,8 +3,33 @@ import { faker } from '@faker-js/faker';
...
 import { TierIds } from '../../shared/constants.js';
 
-// in a terminal window run `wasp db seed` to seed your dev database with mock user data
-export function createRandomUser() {
+/**
+ * This function, which we've imported in `app.db.seeds` in the `main.wasp` file, 
+ * seeds the database with mock users via the `wasp db seed` command.
+ * For more info see: https://wasp-lang.dev/docs/data-model/backends#seeding-the-database
+ */
+export async function seedMockUsersToDB(prismaClient: PrismaClient) {
```
Martinsos, line 11: `seedMockUsersToDb` -> this name is correct and descriptive, so that is great! I would however probably drop `ToDb`, it is clear from context that is what is happening (seeding is term related to db, and file is called `dbSeeds`).

Martinsos (commented):
Great effort here, this will end up being very nice! Left some Clean Code (TM) comments!

template/app/src/server/scripts/dbSeeds.ts
```diff
@@ -3,8 +3,33 @@ import { faker } from '@faker-js/faker';
...
+  const mockUsers: Omit<User, 'id'>[] = getMockUsers(50);
+  const persistUsersToDB = mockUsers.map((user) => {
+    return prismaClient.user.create({
+      data: user,
+    });
+  });
+
+  try {
+    await Promise.all(persistUsersToDB);
+  } catch (error) {
```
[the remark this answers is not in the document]
vincanger: ok. so you don't think this Promise.all() needs to be wrapped in trycatch? I was under the impression promises should be wrapped in trycatch, but I didn't take into account that Prisma might catch the error. I'll purposefully have it throw an error and see :)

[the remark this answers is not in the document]
Martinsos: So Promise.all might throw an error (it will propagae an error from underlying promises).
If that happens, that error will travel upwards and stop stuff till it hits the final boundary, which is for normally `node` if there is nobody else that already caught it and stopped it. Then `node` will print that error. Now, we do have one layer extra, since this is being run by `Prisma` as a seed, so I don't know if Prisma will do some handling of its own. Maybe, maybe not, I don't know.

template/app/src/server/scripts/dbSeeds.ts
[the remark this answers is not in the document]
vincanger: hm. in the end TS wasn't happy that createUser's argument wasn't typed, so I used the `Omit<User, 'id'>` there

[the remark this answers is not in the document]
Martinsos: Aha got it! Alternatively you can inline it in `Promise.all`.

template/app/src/server/scripts/dbSeeds.ts
```diff
@@ -0,0 +1,40 @@
...
+function generateMockUsersData(numOfUsers: number) {
+  return faker.helpers.multiple(generateMockUserData, { count: numOfUsers });
+}
+
+function generateMockUserData() {
+  const firstName = faker.person.firstName();
+  const lastName = faker.person.lastName();
+  const now = new Date();
+  const recently = faker.date.recent({ refDate: now }); // Creates a random date up to 24 hrs ago from refDate
+  const past = faker.date.past({ refDate: now }); // Creates a random date in past up to 1 year from refDate
```
vincanger, line 24: tried my best to name these dates. `past` and `recently` just seemed better than `someDateWithinLastYear`, etc

template/app/src/server/scripts/dbSeeds.ts
```diff
@@ -0,0 +1,40 @@
...
+import type { PrismaClient } from '@prisma/client';
+import { TierIds } from '../../shared/constants.js';
+
+/**
+ * This function, which we've imported in `app.db.seeds` in the `main.wasp` file,
+ * seeds the database with mock users via the `wasp db seed` command.
+ * For more info see: https://wasp-lang.dev/docs/data-model/backends#seeding-the-database
+ */
+export async function seedMockUsers(prismaClient: PrismaClient) {
+  await Promise.all(generateMockUsersData(50).map((user) => prismaClient.user.create({ data: user })));
```
Martinsos, line 12: Nitpick: use `data` instead of `user` both to shorten it and to make it a tidbit more semantically correct.
Btw I am a bit divided on this one if it should be one line or not -> but at the end doesn't really matter much, so both is ok.

```suggestion
  await Promise.all(generateMockUsersData(50).map(
    data => prismaClient.user.create({ data })
  ));
```

template/app/src/server/scripts/dbSeeds.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+function generateMockUsersData(numOfUsers: number) {
```
Martinsos, line 15: In Typescript, it is nice to have return types for fnuctions when possible. It makes it clear what is the API of the function and what it should return. So let's add return type here, and also on the function below.

template/app/src/server/scripts/dbSeeds.ts
Martinsos, line 24: you could do: recentPast, furtherPast or something like that? This is also not terrible. But yeah if you could get away with comments hm. What about `momentInLastDay` and `momentInLastYear`?

Btw why not just use the more directly hm. Let's take a look at how you use them.

You say `createdAt: past`. Ok, fine.

Then you say `lastActiveTimestamp: recently`. Ok, but we said lastActiveTimestamp should be after createdAt. Have you ensured that? I don't think you did, `past` can in theory still be newer than `recently`, right? Not very likely, but I haven't seen it having a minimum on how fresh it can be. So we still have the same issue.

You then set `datePaid` to be in between `past` and `recently`.

Ok, so problem you have is that you did one level too much of indirection here and got messed up with it.

Why don't you do this, since I think this is what you really want:
```js
const createdAt = faker.date.past({ refDate: now });
const lastActiveTimestamp = faker.date.between({ from: createdAt, to: now });  // Here you can maybe also tryn using `recently` instead if it has option to set minimum/maximum/whatever.
const datePaid = faker.date.between({ from: createdAt, to: lastActiveTimestamp });
```

vincanger: yep. makes a lot of sense.

template/app/main.wasp
```diff
@@ -100,8 +100,8 @@ entity User {=psl
 
   stripeId                  String?
   checkoutSessionId         String?
-  subscriptionTier          String?
-  subscriptionStatus        String?
+  subscriptionStatus        String?         // 'active', 'canceled', 'past_due', 'deleted', null
+  subscriptionTier          String?         // 'hobby-tier', 'pro-tier', null
```
Martinsos, line 104: Keeping this on purpose or is this a leftover?

Martinsos (approved):

template/app/main.wasp
vincanger: i'll leave it there until we get enums in v0.14

PR #204, opened 2024-06-29
Update guided-tour.md

fmaiabatista:
## Description

> Describe your PR! If it fixes specific issue, mention it with "Fixes # (issue)".

Hey 👋 

I'm going through the Guided Tour and I found that when I changed my app name, I also had to migrate the DB again, and my previous data was lost. I added this info to the warning box so other users are aware of the impact.

By the way, the `Edit page` button in the docs are pointing to the `deployed-version` branch, which I found to be outdated. I had to find the new path to the doc file in the `main` branch to propose the changes. See below:

image

Let me know if my understanding and proposed changes are accurate and feel free to merge the PR as is or further change the text!

## Contributor Checklist

> Make sure to do the following steps if they are applicable to your PR:

- [x] **Update e2e tests**: If you changed the [/template/app](/template/app), then make sure to do any neccessary updates to [/template/e2e-tests](/template/e2e-tests) also.
- [x] **Update demo app**: If you changed the [/template/app](/template/app), then make sure to do any neccessary updates to [/opensaas-sh/app_diff](/opensaas-sh/app_diff) also. Check [/opensaas-sh/README.md](/opensaas-sh/README.md) for details.
- [x] **Update docs**: If needed, update the [/opensaas-sh/blog/src/content/docs](/opensaas-sh/blog/src/content/docs).

PR #205, opened 2024-07-01
Updated edit page link to point to new repo structure.

Martinsos:
Motivated by https://github.com/wasp-lang/open-saas/pull/204 -> @vincanger , what should we do with the edit page link? It shouldn't point to dpeloyed-version anymore -> so it can either point to `main`, which is good because users edit the latest version, but also a bit confusing because it might be different then what is deployed, or, we have it point to the latest released version, which we would probably have to do by refereing it via git tag (e.g. `template-version-0.13`, which we need to then keep updating in astro.config.js and are likely to forget.

I would maybe just have it point to `main` for the sake of simplicity for now, and once we have a bit more sophisticated way of tracking what is released (e.g. a `release` branch) then we can do something better. If so, this PR is it.

### DIFF
--- a/opensaas-sh/blog/astro.config.mjs
+++ b/opensaas-sh/blog/astro.config.mjs
@@ -36,7 +36,7 @@ export default defineConfig({
         },
       ],
       editLink: {
-        baseUrl: 'https://github.com/wasp-lang/open-saas/edit/deployed-version/blog',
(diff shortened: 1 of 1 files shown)


vincanger:
yeah @Martinsos `main` seems to be the best for now

vincanger (approved):

Issue #201, opened 2024-06-28
Check demo app in CI (patching)


Martinsos:
In CI, we should run patching (opensaas-sh/tools/patch.sh) to confirm that demo app patches correctly based on template and app_diff. If not, CI should fail.

PR #206, opened 2024-07-01
Added check for demo app patching to CI.

Martinsos:
Resolves https://github.com/wasp-lang/open-saas/issues/201

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -26,11 +26,19 @@ jobs:
         with:
           node-version: '20'
 
+      - name: Docker setup
+        uses: docker/setup-buildx-action@v3
+
       - name: Install Wasp
         run: curl -sSL https://get.wasp-lang.dev/installer.sh | sh -s -- -v ${{ env.WASP_VERSION }}
 
-      - name: Docker setup
-        uses: docker/setup-buildx-action@v3
+      - name: Cache global node modules
+        uses: actions/cache@v4
+        with:
+          path: ~/.npm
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -26,11 +26,19 @@ jobs:
         with:
           node-version: '20'
 
+      - name: Docker setup
```
Martinsos, line 29: I moved this and some other steps to make it more like "first setup", then "check patching", then "run e2e tests".

.github/workflows/e2e-tests.yml
```diff
@@ -43,44 +51,41 @@ jobs:
...
           cp .env.server.example .env.server && cp .env.client.example .env.client
 
-      - name: Cache global node modules
-        uses: actions/cache@v4
-        with:
-          path: ~/.npm
-          key: node-modules-${{ runner.os }}-${{ hashFiles('template/app/package-lock.json') }}-${{ hashFiles('template/e2e-tests/package-lock.json') }}-wasp${{ env.WASP_VERSION }}-node${{ steps.setup-node.outputs.node-version }}
-          restore-keys: |
-            node-modules-${{ runner.os }}-
+      - name: Check that patching demo app (opensaas-sh) works
```
Martinsos, line 54: This is the step I added!

opensaas-sh/tools/dope.sh
```diff
@@ -123,6 +126,14 @@ recreate_derived_dir() {
     fi
 
     echo "DONE: generated ${DERIVED_DIR}/"
+
+    if [ ${num_patches_failed} -gt 0 ]; then
+        echo -e "${RED_COLOR}${num_patches_failed} patches failed, look into generated files for merge conflicts.${RESET_COLOR}"
+        exit 1;
```
Martinsos, line 132: I added this so that we return 1 as exit status/code if any of the patches fails, so that CI also fails then.

vincanger (approved):
ok looks good. tested it out locally. I'm going to merge it now and fix the diff :)

PR #207, opened 2024-07-02
create updated app_diff

vincanger:
resolves merge conflicts in opensaas_sh/app and updates app_diff


opensaas-sh/app_diff/main.wasp.diff
```diff
@@ -84,8 +84,12 @@
...
    },
  }
-@@ -93,10 +91,13 @@
-   id                        Int             @id @default(autoincrement())
+@@ -90,13 +88,16 @@
+  */
+ 
+ entity User {=psl
+-  id                        String          @id @default(uuid())
++  id                        Int             @id @default(autoincrement())
```
Martinsos, line 92: We don't want really to switch to Int though, I think we also want to use String now that that change happened on the `main`, right?

opensaas-sh/app_diff/main.wasp.diff
```diff
@@ -101,3 +105,39 @@
  
    stripeId                  String?
    checkoutSessionId         String?
+@@ -118,7 +119,7 @@
+   updatedAt                 DateTime        @updatedAt
+ 
+   user                      User            @relation(fields: [userId], references: [id])
+-  userId                    String
++  userId                    Int
```
Martinsos, line 113: This and changes bleow, I think we also don't want these really.

opensaas-sh/app_diff/src/server/actions.ts.diff
```diff
@@ -1,9 +1,42 @@
...
++      model: 'gpt-3.5-turbo', // you can use any model here, e.g. 'gpt-3.5-turbo', 'gpt-4', etc.
+       messages: [
+         {
+           role: 'system',
+@@ -286,7 +291,7 @@
+   return task;
+ };
+ 
+-export const updateUserById: UpdateUserById<{ id: string; data: Partial<User> }, User> = async (
++export const updateUserById: UpdateUserById<{ id: number; data: Partial<User> }, User> = async (
```
Martinsos, line 31: user id is also number here, should probably be string

opensaas-sh/app_diff/src/server/actions.ts.diff
```diff
@@ -1,9 +1,42 @@
...
++    ) {
+       throw new HttpError(402, 'User has not paid or is out of credits');
+     } else if (context.user.credits && !context.user.subscriptionStatus) {
+       console.log('decrementing credits');
+@@ -128,7 +133,7 @@
+     }
+ 
+     const completion = await openai.chat.completions.create({
+-      model: 'gpt-3.5-turbo', // you can use any model here, e.g. 'gpt-3.5-turbo', 'gpt-4', etc. 
++      model: 'gpt-3.5-turbo', // you can use any model here, e.g. 'gpt-3.5-turbo', 'gpt-4', etc.
```
Martinsos, line 22: Doesn't seem to be a meaningful change, just space at the end.

opensaas-sh/app_diff/src/server/actions.ts.diff
```diff
@@ -1,9 +1,42 @@
 --- template/app/src/server/actions.ts
 +++ opensaas-sh/app/src/server/actions.ts
-@@ -318,6 +318,18 @@
+@@ -113,7 +113,12 @@
+       throw openai;
+     }
+ 
+-    if (!context.user.credits && (!context.user.subscriptionStatus || context.user.subscriptionStatus === 'deleted' || context.user.subscriptionStatus === 'past_due')) {
++    if (
```
Martinsos, line 8: Just formatting, probably we don't want to do it here.

opensaas-sh/app_diff/src/server/actions.ts.diff
```diff
@@ -16,6 +49,7 @@
 +    throw new HttpError(403, 'Thanks for trying Open SaaS. This demo only allows 2 file uploads per user.');
 +  }
 +
-   const userInfo = context.user.id.toString();
++  const userInfo = context.user.id.toString();
```
Martinsos, line 52: If we go for using string for user id (and we should), then we can also drop the toString part here.

Martinsos (changes requested):

opensaas-sh/app_diff/main.wasp.diff
vincanger: well this will cause issues in current app and erase all the nice data available in its admin dashboard

Martinsos: Yeah you are right, unless we do some data migration but that is a bit much work now. But on the other hand, I love this idea where demo app is just minimal additions on top of teh template, and this is not it anymore, this is legacy stuff. So I am tempted to just drop the data we have. Will people care at all? They are expecting mock data anyway. This will not be the only such situation btw, we need to know how we are handling situations like this. I am leaning on the side of just resetting the db whenever it suits us, it is a demo app and it is the fastest way.

Martinsos: Oh nice, great!

Martinsos (approved):
Great, approved!

PR #217, opened 2024-07-04
Martin's attempt at consolidating types.

Martinsos:
@vincanger I wanted to see if some of the suggestions I have actually stand, so I tried experimenting, and ended up implementing something at the end. Take a look if there are parts here that you find useful. I haven't tried running it, but it compiles, and I might have missed something somewhere, but basic idea should be here. And I haven't fixed all the issues from origianl PR, just tried to get these new types going.

The main shortcoming here is that I removed the difference between the subscription and credits at the type level, in an attempt to simplify things. I don't think result is so bad in practice, but it depends on what you want I guess. Might be worth trying to add that distinction back, maybe.


template/app/src/payment/plans.ts
Martinsos, line 1: I centralized stuff in this file.

template/app/src/payment/plans.ts
```diff
@@ -0,0 +1,51 @@
...
+  stripePriceId: string,
+  effect: PaymentPlanEffect
+}
+
+export type PaymentPlanEffect = { kind: 'subscription' } | { kind: 'credits', amount: number }
+export type PaymentPlanEffectKinds = PaymentPlanEffect extends { kind: infer K } ? K : never;
+
+export const paymentPlans: Record<PaymentPlanId, PaymentPlan> = {
+  [PaymentPlanId.Hobby]: {
+    stripePriceId: requireNodeEnvVar('STRIPE_HOBBY_SUBSCRIPTION_PRICE_ID'),
```
Martinsos, line 18: Might this be problem since it assumes we are running in node? And this file is intended to also be used on client! Yeah, maybe we should make this afunction instead (`getStripePriceId: () => requireNodeEnvVar(...`), that might be enough. Or if not, separate it into another file.

template/app/src/payment/plans.ts
```diff
@@ -0,0 +1,51 @@
...
+  Credits10 = 'credits10'
+}
+
+export interface PaymentPlan {
+  getStripePriceId: () => string,
+  effect: PaymentPlanEffect
+}
+
+export type PaymentPlanEffect = { kind: 'subscription' } | { kind: 'credits', amount: number };
+export type PaymentPlanEffectKinds = PaymentPlanEffect extends { kind: infer K } ? K : never;
```
sodic, line 14:
```typescript
export type PaymentPlanEffectKinds = PaymentPlanEffect['kind']
```

PR #221, opened 2024-07-08
Fix MailGun in main.wasp to Mailgun

Martinsos:


vincanger:
@Martinsos you have to update the opensaas-sh app_diff as there is a merge conflict

Martinsos:
> @Martinsos you have to update the opensaas-sh app_diff as there is a merge conflict

Thanks, handled!

PR #244, opened 2024-07-25
S3 upload custom endpoint url

peppelauro:
## Description

> Allow to use also a non-Amazon S3 provider like OVH, DigitalOcean and so on

## Contributor Checklist


vincanger:
Thanks for the contribution here @peppelauro. What happens if the user wants to use Amazon S3 and not a custome endpoint or 3rd-party provider? Shouldn't we add a conditional check and maybe some comments in the files to explain what's going on? What do you think?

vincanger (changes requested):
Thanks for the contribution here @peppelauro. What happens if the user wants to use Amazon S3 and not a custom endpoint or 3rd-party provider? Could this possible be confusing for them? Shouldn't we add some comments in the files or docs to explain what's going on? What do you think?

peppelauro:
And of course, in case you validate the previous snippet, we will add a commented line in the .env.server:

`
# (OPTIONAL) get your aws s3 credentials at https://console.aws.amazon.com and create a new IAM user with S3 access
#If you want to use a non-amazon S3 decomment the following line:
#AWS_S3_ENDPOINT_URL=https://...
AWS_S3_IAM_ACCESS_KEY=ACK...
AWS_S3_IAM_SECRET_KEY=t+33a...
AWS_S3_FILES_BUCKET=your-bucket-name
AWS_S3_REGION=your-region
`

vincanger:
@peppelauro on second thought, I think it's too confusing to add the conditional code and optional env variable to the template. I think it's best if we don't clutter it too much. We can only offer so many integrations and examples for the user, so if the user wants to use something else besides AWS S3, that's up to them to figure out how to integrate it.

With that said, what do you think about adding a section to the "File Upload" docs, e.g. "Using a non-Amazon S3 provider (OVH, DigitalOcean)" that includes this information and code snippet?

PR #246, opened 2024-08-02
add lemon squeezy as payment provider

### LINKED ISSUE #16: add lemonsqueezy payment support
Open SaaS currently has subscription payments via Stripe.

It would be nice to add similar support for [Lemon Squeezy](https://www.lemonsqueezy.com/) as mentioned by @shengxinjing

### Process

- move the action `stripePayment` from `app/src/server/actions.ts` to `app/src/server/stripeUtils.ts`
- create a similar file for lemonsqueezy, e.g. `app/src/server/lemonsqueezyUtils.ts` and add the complimentary checkout logic for lemon squeezy. If possible, we want to make the checkout action interchangeable with the `stripePayment` action, so that these actions can swapped 1:1 on the client
- create a complimentary webhook file in `app/src/server/webhooks` if necessary.
- update `app/src/client/app/PricingPage.tsx` if necessary

### LINKED ISSUE #103: Can we integrate other payment gateways
I would like to know if it is possible to integrate other local payment gateways than stripe ?

vincanger:
Resolves #103  and Resolves #16 

- [x] Add lemonsqueezy checkout (operation, util functions, etc)
- [x] Add webhook and handle necessary events
- [x] Add operations to pricing page 
- [x] Add ability to access lemonsqueezy customer portal after user pays 
    - [x] on Pricing Page
    - [x]  on Account Page
- [x] Test one-time payment product
- [x] add lemonsqueezy analytics methods
- [x] update app_diff
- [x] Add docs

when deploying, make sure to:
- [ ] change stripe webhook url `payments-webhook`
- [ ] update product plan id env vars

Note: when developing with webhooks locally, use [ngrok](https://ngrok.com/) and create a tunnel to server on port 3001. Add the url it provides you + `/lemonsqueezy-webhook` to your webhook settings on the lemonsqueezy dashboard e.g. `9bbf-176-199-209-145.ngrok-free.app/lemonsqueezy-webhook`:

```sh
ngrok http 3001
```


opensaas-sh/blog/src/content/docs/guides/lemon-squeezy-integration.md
```diff
@@ -0,0 +1,116 @@
...
+Once installed, and with your wasp app running, run:
+```sh
+ngrok http 3001
+```
+
+The Wasp server will be running on port 3001, so this creates an accessbile tunnel to your server.
+
+![ngrok](/lemon-squeezy/ngrok.png)
+
+Ngrok will output a forwarding address for you. Copy and paste this address and add `/lemonsqueezy-webhook` to the end (this URL path has been configured for you already in `main.wasp` under the `api lemonSqueezyWebhook` definiton). It should look something like this:
```
Martinsos, line 65:
```suggestion
Ngrok will output a forwarding address for you. Copy and paste this address and add `/lemonsqueezy-webhook` to the end (this URL path has been configured for you already in `main.wasp` under the `api lemonSqueezyWebhook` definition). It should look something like this: 
```

There are probably other small spelling mistakes, maybe let's run it through grammarly or GPT?

Martinsos, line 65: Aaaah ok, so basically ngrok will make my locally running Wasp app available online (well, its server). Right, I forgot for a moment that is what ngrok does.

So let's maybe explain it that way all together? Smth like "Lemonsquezy sends messages/updates to your Wasp app (the webhook). For that to work during development, we need to expose our locally running (via `wasp start`) Wasp app and make it available online, specifically the server part of it. Since Wasp server runs on port 3001, you should run `ngrok 3001`, which will provide you with a public URL that you can then ... "

opensaas-sh/blog/src/content/docs/guides/lemon-squeezy-integration.md
```diff
@@ -0,0 +1,116 @@
...
+  - order_created
+  - subscription_created
+  - subscription_updated
+  - subscription_cancelled
+- click `save`
+
+You're now ready to start consuming Lemon Squeezy webhook events in local development.
+
+:::caution[Restart the Tunnel after Webhook changes]
+If you make any changes to your Webhook code in `payment/lemonSqueezy/webhook.ts`, you'll need to restart ngrok for these changes to take effect via the tunnel.
```
Martinsos, line 86: Oh really? Weird, I find that unexpected, how is that? Isn't it just forwarding stuff from that public URL to port 3001 on the machine? if I update webhook.ts, I would expect that Wasp's node server restarts, but keeps running on 3001 and ngrok doesn't have to know anything about that.

opensaas-sh/blog/src/content/docs/guides/lemon-squeezy-integration.md
```diff
@@ -0,0 +1,116 @@
...
+:::
+
+## Update your App code to use Lemon Squeezy instead of Stripe
+
+There are a few places in your Wasp app where we need to switch out Stripe logic for Lemon Squeezy logic. So that you can easily find them, we've left the comment `// PAYMENTS PROCESSOR:` in the spots where you need to switch out the Stripe logic for Lemon Squeezy's. You can easily search for these comments, remove the Stripe functions, and uncomment out the Lemon Squeezy function, e.g.:
```
Martinsos, line 91: It would be great if we could make this even simpler. Hunting down these comments and updating them is quite tricky, also makes testing hard for us. Feels wrong.

Typical programming pattern for stuff like this is called Strategy (https://en.wikipedia.org/wiki/Strategy_pattern). Although Strategy is more about choosing the algorithm during runtime, while we probably want them to choose it during configuring the project (delete code you don't need and keep the rest)? So that might be closer to Facade. I forgot my design patterns a bit hah, but essentially, what you ideally want is to define a single interface called `PaymentProcessor` for example, that defines all the stuff that a payment processor has to provide (`generateCheckoutSession` function, ...), and then you would have stripePaymentProcessor.ts and lemonSqueezyPaymentProcessor.ts and then you would also have `paymentProcessor.ts` where you would have `const paymentProcessor = stripePaymentProcessor; export paymentProcessor` (or lemon squeezy) for example and the rest of the code imports `paymentProcessor` and doesn't have to know which one is it exactly. The point is that there is a single place where this decision is made.

The question is, how well can you extract it like that, sometimes in practice it can get a bit tricky (env vars, some special cases, ...), but I think if we try we will get a good result.

Then they just have to delete one file and make sure in that  single line correct processor is set, there is no hunting down the lines with comments.

You have two impelmentations: Stripe or LemonSqueezy, and you want to define this common interface for both, behind which they will be hiding and used.

template/app/.prettierrc
Martinsos, line 1: Aha so they don't get prettierrc out of the box anymore, but instead we have one for ourselves.

template/app/main.wasp
```diff
@@ -200,12 +200,24 @@ action generateStripeCheckoutSession {
   entities: [User]
 }
 
+action generateLemonSqueezyCheckoutSession {
```
Martinsos, line 203: Ok, so if you go with the interfaces and implementations approach that I described, then you ideally wouldn't have separate ops here for Stripe and LemonSqueezy, you would have just one set of these for "payment processor" as a general concept

template/app/package.json
```diff
@@ -6,6 +6,7 @@
     "@faker-js/faker": "8.3.1",
     "@google-analytics/data": "4.1.0",
     "@headlessui/react": "1.7.13",
+    "@lemonsqueezy/lemonsqueezy.js": "^3.2.0",
```
Martinsos, line 9: Ok, we do need to warn them that they should delete this if htey are not using lemonsqueezy

template/app/schema.prisma
```diff
@@ -17,6 +17,8 @@ model User {
   isAdmin                   Boolean         @default(false)
 
   stripeId                  String?         @unique
+  lemonSqueezyId            String?         @unique
```
Martinsos, line 20: Hm yeah this will be a bit trickier to "modularize". How is it that lemon squeezy also needs this customer portal url and stripe doesn't? Is there any way to make this stuff not specific to processor, but unified? So we have let's say just one field called `paymentProcessorId`?

template/app/src/payment/lemonSqueezy/checkoutUtils.ts
```diff
@@ -0,0 +1,37 @@
+import { createCheckout } from '@lemonsqueezy/lemonsqueezy.js';
+import { initLemonSqueezy } from './initLemonSqueezySetup';
+
+initLemonSqueezy();
```
Martinsos, line 4: A bit weird to see this here, I wonder if this is ok. Why here, why not in some other file?

I would expect this to be handled in one of two ways:
1. You let "user" of the lemon squeezy code call this on their own. Lemon squezy code here being everything under `payment/lemonSqueezy` directory, and "user" being the rest of open-saas code. So you would somewhere in open-saas code call this (maybe server setup?).
2. Or, you put it in some file in `payment/lemonSqueezy` that is somehow central to the whole thing. For example, maybe `index.js`, but then you are probably also expecting most of the stuff for this module tree to be exposed via `index.js`, so it serves as a central entrypoint. Basically, it needs to be a file they need to import. And I do get that they need to import all of these files, but they are all of same improtance and that makes it problematic, due to possible different import orders and similar. Even if not problematic in practice, it feels wrong. If you go with interface and implementation design pattern, then you will proably have such central file so that will be a good place to do this intilaization.

template/app/src/payment/lemonSqueezy/checkoutUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+type LemonSqueezyCheckoutSessionParams = {
```
Martinsos, line 6: Why not interface? By default, it is recommended to use interface over type if both fit.

Martinsos (commented):
@vincanger awesome to see this working! This is looking great in general, having all the logic for wokring with lemon squeezy. I think the main thing we can improve is just how it is shaped, in the sense that we coudl try modularizing both it and Stripe to be easily interchangeable and therefore easy to remove/pick. I mostly comented on that. I also haven't checked all the details of lemonsqueezy logic, but I can check that once we do the modularization.

.github/workflows/e2e-tests.yml
```diff
@@ -113,6 +113,7 @@ jobs:
           STRIPE_CREDITS_PRICE_ID: ${{ secrets.STRIPE_CREDITS_PRICE_ID }}
           SKIP_EMAIL_VERIFICATION_IN_DEV: true
           # Client-side env vars
+          REACT_APP_STRIPE_CUSTOMER_PORTAL: https://billing.stripe.com/p/login/test_8wM8x17JN7DT4zC000
```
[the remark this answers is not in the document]
vincanger: the dummy url in .env.example was the same so it was fulfilling the regex check.

opensaas-sh/app_diff/src/landing-page/contentSections.ts.diff
```diff
@@ -44,8 +44,8 @@
    {
 -    name: 'Cool Feature #4',
 -    description: 'Describe your cool feature here.',
-+    name: 'Stripe Integration',
-+    description: "No SaaS is complete without payments. That's why payments and the necessary webhooks are built-in.",
++    name: 'Stripe / Lemon Squeezy Integration',
++    description: "No SaaS is complete without payments. We've pre-configured checkout and webhooks. Just choose a provider and start cashing out.",
      icon: '💸',
 +    href: DocsUrl + '/guides/stripe-integration/',
```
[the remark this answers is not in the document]
vincanger: Hmm yeah. Maybe I can make a "payments" section with these as sub-sections

template/app/schema.prisma
vincanger: Lemon Squeezy gives each user a specific customer portal URL, while Stripe uses a general login portal for your store. I'm not sure how to best generalize this into one variable though...

aiddroid:
Please reactor this into a common payment interface.

opensaas-sh/blog/src/content/docs/guides/lemon-squeezy-integration.md
vincanger: Ok @Martinsos this makes total sense. I took a crack at this and now there is a central `paymentProcessor` file. Could you take a look and give me some feedback before I continue with the rest of the PR? This way I know if I'm going in the right direction. Thanks!

template/app/src/payment/paymentProcessor.ts
```diff
@@ -0,0 +1,14 @@
...
+import { PrismaClient } from '@prisma/client';
+import { stripePaymentProcessor } from './stripe/paymentProcessor';
+import { lemonSqueezyPaymentProcessor } from './lemonSqueezy/paymentProcessor';
+
+export interface CreateCheckoutSessionArgs { userId: string | undefined; userEmail: string; paymentPlan: PaymentPlan }
+export interface FetchCustomerPortalUrlArgs { userId: string; prismaUserDelegate: PrismaClient['user'] };
+
+interface PaymentProcessor {
+  createCheckoutSession: ({ userId, userEmail, paymentPlan }: CreateCheckoutSessionArgs) => Promise<{ session: { id: string; url: string }; customer?: { id: string } }>;
+  fetchCustomerPortalUrl: (args?: FetchCustomerPortalUrlArgs) => Promise<string>;
```
vincanger, line 11: I also took a crack at solving our customer portal URL issue, and making it more unified. I tried some different approaches here when creating the type, and although this one works, I am not really happy with how it functions.

The issue here is that for Stripe, we have only one customer portal URL per store, than any customer can log in to, whereas with Lemon Squeezy (LS), each customer has a unique URL assigned to them, which we receive in the webhook on checkout. To fetch the Stripe URL, no arguments are needed, but for the LS URL, we need to search for it by USER

opensaas-sh/app_diff/src/landing-page/contentSections.ts.diff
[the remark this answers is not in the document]
Martinsos: Yup I think that would make sense!

template/app/src/payment/lemonSqueezy/paymentDetails.ts
```diff
@@ -0,0 +1,32 @@
+import type { SubscriptionStatus } from '../plans';
+import { PaymentPlanId } from '../plans';
+import { PrismaClient } from '@prisma/client';
+
+type UserLemonSqueezyPaymentDetails = {
```
Martinsos, line 5: Probably should be interface, it is rcommended to use interface instead of type when possible.

template/app/src/payment/operations.ts
```diff
@@ -47,10 +38,12 @@ export const generateStripeCheckoutSession: GenerateStripeCheckoutSession<
...
 
-function paymentPlanEffectToStripeMode(planEffect: PaymentPlanEffect): StripeMode {
-  const effectToMode: Record<PaymentPlanEffect['kind'], StripeMode> = {
-    subscription: 'subscription',
-    credits: 'payment',
-  };
-  return effectToMode[planEffect.kind];
-}
+export const getCustomerPortalUrl: GetCustomerPortalUrl<void, string> =  async (_args, context) => {
+    if (!context.user) {
```
Martinsos, line 42: Someothing wrong with indentation?

template/app/src/payment/paymentProcessor.ts
```diff
@@ -0,0 +1,14 @@
+import type { PaymentPlan } from './plans';
+import { PrismaClient } from '@prisma/client';
+import { stripePaymentProcessor } from './stripe/paymentProcessor';
+import { lemonSqueezyPaymentProcessor } from './lemonSqueezy/paymentProcessor';
+
+export interface CreateCheckoutSessionArgs { userId: string | undefined; userEmail: string; paymentPlan: PaymentPlan }
```
Martinsos, line 6: I am not TS master, so maybe we should ask Filip for opinion here once we are ready, but my first insinct is to have only `interface PaymentProcessor` and then if I need the args of one of the functions, I would do `Parameters<PaymentProcessor["createCheckoutSession"]>[0]` or sometihgn like that. But maybe I am wrong here. We can figure this out a bit later though, it is a minor thing really.

template/app/src/payment/paymentProcessor.ts
```diff
@@ -0,0 +1,14 @@
...
+export interface FetchCustomerPortalUrlArgs { userId: string; prismaUserDelegate: PrismaClient['user'] };
+
+interface PaymentProcessor {
+  createCheckoutSession: ({ userId, userEmail, paymentPlan }: CreateCheckoutSessionArgs) => Promise<{ session: { id: string; url: string }; customer?: { id: string } }>;
```
Martinsos, line 10: Interesting, they both return same type for a session? Both stripe and lemon squeezy?

template/app/src/payment/paymentProcessor.ts
Martinsos, line 6: Why can `userId` be `undefined`? Because one of them doesn't need it? How about we make it required and they just don't have to use it?

template/app/src/payment/paymentProcessor.ts
Martinsos: How about we require the arguments, but then we don't use them for Stripe and that is it?

template/app/schema.prisma
```diff
@@ -16,7 +16,8 @@ model User {
   lastActiveTimestamp       DateTime        @default(now())
   isAdmin                   Boolean         @default(false)
 
-  stripeId                  String?         @unique
+  paymentProcessorId        String?         @unique
```
Martinsos, line 19: This is really a user id up there on that payment processor right?
Would it make ense to call this `paymentProcessorUserId`? or do you think it is enough context that this is already in the User entity? It is a bit hard to say for me but you have more context.

template/app/schema.prisma
```diff
@@ -16,7 +16,8 @@ model User {
...
+  lemonSqueezyCustomerPortalUrl String?
```
Martinsos, line 20: Yeah this is a bit tricky because lemon squeezy needs this and stripe doesn't.
We could in theory introduce two new entities, one called `LemonSqueezyPaymentDetails` and another called `StripePaymentDetails` and then have those link to the User, but I am not sure if we want to go that far at the moment or if we would rather keep things simple.

So this here is fine then I guess! We could in theory rename it to `customerPortalUrl` only potentially. But maybe being specific is also fine.

template/app/src/payment/stripe/paymentProcessor.ts
```diff
@@ -0,0 +1,32 @@
...
+      priceId: paymentPlan.getPriceId(),
+      customerId: customer.id,
+      mode: paymentPlanEffectToStripeMode(paymentPlan.effect),
+    });
+    if (!stripeSession.url) throw new Error('Error creating Stripe Checkout Session');
+    const session = {
+      url: stripeSession.url,
+      id: stripeSession.id,
+    };
+    return { session, customer };
```
Martinsos, line 21: Why do we also return customer in this case, but not in the case of lemon squeezy?

template/app/src/payment/stripe/stripeClient.ts
```diff
@@ -1,11 +1,12 @@
 import Stripe from 'stripe';
+import { requireNodeEnvVar } from '../../server/utils';
 
-export const stripe = new Stripe(process.env.STRIPE_KEY!, {
+export const stripe = new Stripe(requireNodeEnvVar('PAYMENTS_API_KEY'), {
```
Martinsos, line 4: So now we have PAYMENTS_API_KEY for both.
Hmmm that is aligned with the idea of unification, but I wonder if that is a step too far. I think having separated STRIPE_KEY and LEMON_SQUEEZY_KEY might be just all right. It is kind of an implementation detail, and it keeps things clear, when we have both and they are separated.

template/app/src/payment/stripe/webhook.ts
```diff
@@ -12,7 +12,7 @@ import { requireNodeEnvVar } from '../../server/utils';
 import { z } from 'zod';
 
 export const stripeWebhook: StripeWebhook = async (request, response, context) => {
-  const secret = requireNodeEnvVar('STRIPE_WEBHOOK_SECRET');
+  const secret = requireNodeEnvVar('PAYMENTS_WEBHOOK_SECRET');
```
Martinsos, line 15: I wonder also if this env var should be like this or we should maintain two of them.

template/app/src/payment/stripe/webhook.ts
```diff
@@ -86,7 +86,7 @@ export async function handleCheckoutSessionCompleted(
   const lineItemPriceId =  result.data.data[0].price.id;
 
   const planId = Object.values(PaymentPlanId).find(
-    (planId) => paymentPlans[planId].getStripePriceId() === lineItemPriceId
+    (planId) => paymentPlans[planId].getPriceId() === lineItemPriceId
```
Martinsos, line 89: `getPriceId` -> how much sense does `price` here make sense when there is no `stripe` in front of it? Does LS also call it `price`?

Martinsos (commented):
@vincanger seems to me like this is going in right direction, but we need to figure out some small details yet. I hope my comments will help!

template/app/src/payment/paymentProcessor.ts
vincanger: no way, josé

template/app/src/payment/lemonSqueezy/checkoutUtils.ts
```diff
@@ -0,0 +1,29 @@
+import { createCheckout } from '@lemonsqueezy/lemonsqueezy.js';
+
+interface LemonSqueezyCheckoutSessionParams {
```
Martinsos, line 3: Why do you extract this like this, why not just define these directly in the function?
```ts
export async function createLemonSqueezyCheckoutSession(
 { storeId, variantId, userEmail, userId }:
 { storeId: string, variantId: string, userEmail: string, userId: string }
) {
```

This is nicer / easier to read for you? I am ok with it if you like it, it just feels a bit redundant to have to give it a name and I don't see it used anywhere else.

template/app/src/payment/paymentProcessor.ts
```diff
@@ -0,0 +1,32 @@
...
+  prismaUserDelegate: PrismaClient['user']; 
+};
+
+export interface PaymentProcessor {
+  id: 'stripe' | 'lemonsqueezy';
+  createCheckoutSession: ({ userId, userEmail, paymentPlan, prismaUserDelegate }: CreateCheckoutSessionArgs) => Promise<{ session: { id: string; url: string }; }>; 
+  fetchCustomerPortalUrl: ({ userId, prismaUserDelegate }: FetchCustomerPortalUrlArgs) => Promise<string | undefined>;
+  webhook: PaymentsWebhook;
+  webhookMiddlewareConfigFn: MiddlewareConfigFn;
+}
```
Martinsos, line 25:
```suggestion
export interface PaymentProcessor {
  id: 'stripe' | 'lemonsqueezy';
  createCheckoutSession: (args: CreateCheckoutSessionArgs) => Promise<{ session: { id: string; url: string }; }>; 
  fetchCustomerPortalUrl: (args: FetchCustomerPortalUrlArgs) => Promise<string | undefined>;
  webhook: PaymentsWebhook;
  webhookMiddlewareConfigFn: MiddlewareConfigFn;
}
```

template/app/src/payment/paymentProcessor.ts
```diff
@@ -0,0 +1,32 @@
...
+export interface FetchCustomerPortalUrlArgs { 
+  userId: string; 
+  prismaUserDelegate: PrismaClient['user']; 
+};
+
+export interface PaymentProcessor {
+  id: 'stripe' | 'lemonsqueezy';
+  createCheckoutSession: ({ userId, userEmail, paymentPlan, prismaUserDelegate }: CreateCheckoutSessionArgs) => Promise<{ session: { id: string; url: string }; }>; 
+  fetchCustomerPortalUrl: ({ userId, prismaUserDelegate }: FetchCustomerPortalUrlArgs) => Promise<string | undefined>;
```
Martinsos, line 22: How is that return type of `fetchCustomerPortalUrl` is `string | undefined`? Why can it be `undefined`?

template/app/src/payment/plans.ts
```diff
@@ -9,23 +9,23 @@ export enum PaymentPlanId {
 }
 
 export interface PaymentPlan {
-  getStripePriceId: () => string;
+  getPaymentProcessorPlanId: () => string; // ID of the product/plan you've created via your payment processor.
```
Martinsos, line 12: Seemed to me like this could do with some more clarification, does this seem a bit more detailed? I am not sure if the `product id on Stripe` makes sense but you will know how to fix it.

```suggestion
  // Returns the id under which this payment plan is identified on your payment processor. E.g. this might be product id on Stripe.
  getPaymentProcessorPlanId: () => string;
```

template/app/src/payment/stripe/checkoutUtils.ts
```diff
@@ -53,6 +44,9 @@ export async function createStripeCheckoutSession({
         address: 'auto',
       },
       customer: customerId,
+      metadata: {
```
Martinsos, line 47: Aha you did this! Ok cool, so now it also sends it back in the webhook?

template/app/src/payment/lemonSqueezy/webhook.ts
```diff
@@ -0,0 +1,201 @@
...
+      response.status(err.statusCode).json({ error: err.message });
+    } else {
+      response.status(400).json({ error: 'Error Processing Lemon Squeezy Webhook Event' });
+    }
+  }
+};
+
+export const lemonSqueezyMiddlewareConfigFn: MiddlewareConfigFn = (middlewareConfig) => {
+  middlewareConfig.delete('express.json');
+  middlewareConfig.set('express.raw', express.raw({ type: 'application/json' }));
```
Martinsos, line 64: Why is this needed? Looks like we remove json parsing but then add it back? I guess not, but maybe comment would be useful?

template/app/src/payment/lemonSqueezy/webhook.ts
```diff
@@ -0,0 +1,201 @@
...
+  if (!planId) {
+    throw new Error(`No plan with lemonsqueezy variant id ${variantId}`);
+  }
+  const plan = paymentPlans[planId];
+
+  const lemonSqueezyCustomerPortalUrl = await fetchUserCustomerPortalUrl({ lemonSqueezyId });
+
+  let numOfCreditsPurchased: number | undefined = undefined;
+  let datePaid: Date | undefined = undefined;
+  if (status === 'paid' && plan.effect.kind === 'credits') {
```
Martinsos, line 86: Intersting so both has to be true? Status `paid` and also `credits`?

template/app/src/payment/lemonSqueezy/webhook.ts
```diff
@@ -0,0 +1,201 @@
...
+  );
+
+  console.log(`Order ${order_number} created for user ${lemonSqueezyId}`);
+}
+
+async function handleSubscriptionCreated(data: Subscription, userId: string, prismaUserDelegate: PrismaClient['user']) {
+  const { customer_id, status, variant_id } = data.data.attributes;
+  const lemonSqueezyId = customer_id.toString();
+
+  const planId = Object.values(PaymentPlanId).find((planId) => paymentPlans[planId].getPaymentProcessorPlanId() === variant_id.toString());
```
Martinsos, line 103: I believe I already saw the same logic in the function above: should we extract it into a helper function here in this file and then use that so we don't repeat it? It can also throw this error right below it.

template/app/src/payment/lemonSqueezy/webhook.ts
```diff
@@ -0,0 +1,201 @@
...
+    );
+  } else {
+    console.warn(`Unexpected status '${status}' for newly created subscription`);
+  }
+
+  console.log(`Subscription created for user ${lemonSqueezyId}`);
+}
+
+
+// NOTE: LemonSqueezy's 'subscription_updated' event is sent is a catch-all and fires even after 'subscription_created' & 'order_created'
```
Martinsos, line 128:
```suggestion
// NOTE: LemonSqueezy's 'subscription_updated' event is sent as a catch-all and fires even after 'subscription_created' & 'order_created'.
```
I assume

Martinsos: And a period at the end so I know the rest of the sentence wasn't just deleted by accident.

template/app/src/payment/lemonSqueezy/webhook.ts
```diff
@@ -0,0 +1,201 @@
...
+  // NOTE: ability to pause or trial a subscription is something that has to be additionally configured in the lemon squeezy dashboard. 
+  // If you do enable these features, make sure to handle these statuses here.
+  if (status === 'past_due' || status === 'active') {
+    await updateUserLemonSqueezyPaymentDetails(
+      {
+        lemonSqueezyId,
+        userId,
+        subscriptionPlan: planId,
+        subscriptionStatus: status,
+        datePaid: new Date(),
```
Martinsos, line 151: What if status is past_due, is it ok to also udpate datePaid then?

Martinsos (commented):
So many changes in the opensaas-sh demo app! That is quite a lot, I wish we didn't have to diverge that much from the template. Because we really want to keep opensaas-sh as close to the template as possible, as that makes it easy for us to maintain it and update it as template changes. The bigger the difference, the more work for us to maintain it.

What is tricky here is that this is quite a big change through so many files. I guess the diff files help but still!

So the question would be:
1. Can we somehow make the difference smaller?
2. What if we drop this ability to switch between two providers in demo app? Is that really important? We can just write a text saying that they can choose and that is it. What value do they really get by being able to try it out in demo appm, is it worth the complication on our side?

Btw, a separate question: do you expect anybody might want to use two providers at the same time? Stripe AND lemon squeezy? Would that make sense for anybody? That is not typical, right? I am just wondering if we need to think about supporting that or not.

opensaas-sh/app_diff/src/payment/plans.ts.diff
```diff
@@ -0,0 +1,64 @@
...
++import { type PaymentProcessorId } from './paymentProcessor';
+ 
+ export type SubscriptionStatus = 'past_due' | 'cancel_at_period_end' | 'active' | 'deleted';
+ 
+@@ -9,7 +11,7 @@
+ }
+ 
+ export interface PaymentPlan {
+-  getPaymentProcessorPlanId: () => string; // ID of the product/plan you've created via your payment processor. 
++  getPaymentProcessorPlanId: (id: PaymentProcessorId) => string; // ID of the product/plan you've created via your payment processor.
```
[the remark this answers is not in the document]
vincanger: this is removed now

template/app/src/payment/lemonSqueezy/checkoutUtils.ts
vincanger: i find it easier to read personally

template/app/src/payment/paymentProcessor.ts
vincanger: if the user hasn't purchased anything they won't have a URL yet

template/app/src/payment/plans.ts
vincanger: nice. i like it.

template/app/src/payment/stripe/checkoutUtils.ts
vincanger: Ah nice that you caught this. it actually needs to be deleted. I tried it out and it doesn't work the same way as Lemon Squeezy unfortunately

template/app/src/payment/lemonSqueezy/webhook.ts
vincanger: yeah I saw that an order can be created but possibly have an unpaid status

vincanger: image

template/app/src/payment/lemonSqueezy/webhook.ts
vincanger: good catch. now its:

```ts

  if (status === 'past_due' || status === 'active') {
    await updateUserLemonSqueezyPaymentDetails(
      {
        lemonSqueezyId,
        userId,
        subscriptionPlan: planId,
        subscriptionStatus: status,
        datePaid: status === 'past_due' ? undefined : new Date(),
      },
      prismaUserDelegate
    );
```

template/app/src/payment/lemonSqueezy/checkoutUtils.ts
Martinsos: Ok, whatever you like the best then!

template/app/src/payment/lemonSqueezy/webhook.ts
Martinsos: Got it! At which point do we then register that it was paid, if it is not subscription? I guess it won't be any of the subscription events, but how will we know then?

template/app/src/payment/lemonSqueezy/webhook.ts
Martinsos: Got it! Btw this won't set datePaid to undefined in the db, it will just not set it to anything, right?
I think one popular way to write stuff like this is:
```ts
{
  userId,
  ... (status === 'active' && { datePaid: newDate() })
}
```
What I like here is that it is a bit clearer that it is not set at all if status is not `active`. Btw maybe I made slight syntax mistake but it should be close to this.

template/app/src/payment/paymentProcessor.ts
Martinsos: Got it! Maybe write a comment about that? Here, on top of this function.

Martinsos (approved):
@vincanger nice work, I think this is it! Couple of tiny comments left but I am approving, you will solve those as you like, so feel free to merge when happy.
Also this one you missed I think: https://github.com/wasp-lang/open-saas/pull/246#discussion_r1720228408 .

opensaas-sh/blog/src/content/docs/guides/lemon-squeezy-integration.md
vincanger: @Martinsos when I tried it again it worked. I don't know what was causing the issue previously.

template/app/src/payment/lemonSqueezy/webhook.ts
vincanger: @Martinsos yeah, setting to `null` will actually cause a db mutation with Prisma. `undefined` does nothing. 

good tip, but won't it return `false`? could this be an issue with Prisma? I need to look this uop

vincanger: ok I see now that you're using this "optional chaining operator" which only includes it if its true. cool!

template/app/src/payment/lemonSqueezy/webhook.ts
vincanger: For one-time orders, `order_created` is the only event fired. So it seems this would be refired with `paid` once the user pays:

"Lemon Squeezy doesn't seem to have a separate "order_paid" event for one-time orders that were initially unpaid. The "order_created" event serves this purpose, as orders are typically only "created" in the system once payment is confirmed."

PR #251, opened 2024-08-14
Improve updating user lastActiveTimestamp

vincanger:
there was an issue where a user was getting an [infinite loop](https://discord.com/channels/686873244791210014/1272608538400653423) on the `updateCurrentUser` action, most likely because it wasn't being `await`ed on the client side, so I fixed that here.

### DIFF
--- a/template/app/src/client/App.tsx
+++ b/template/app/src/client/App.tsx
@@ -1,11 +1,10 @@
 import './Main.css';
-import { type AuthUser } from 'wasp/auth';
 import { useAuth } from 'wasp/client/auth';
-import { updateCurrentUser } from 'wasp/client/operations';
 import AppNavBar from './components/AppNavBar';
-import CookieConsentBanner from './components/cookie-consent/Banner';
-import { useMemo, useEffect, ReactNode, useState, useCallback } from 'react';
 import { useLocation } from 'react-router-dom';
+import { useMemo, useEffect, ReactNode } from 'react';
+import { updateCurrentUser } from 'wasp/client/operations';
+import CookieConsentBanner from './components/cookie-consent/Banner';
 
 /**
  * use this component to wrap all child components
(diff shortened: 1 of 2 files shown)


driveliveinfo (approved):
Great, thanks!

template/app/src/client/App.tsx
```diff
@@ -14,6 +14,19 @@ export default function App({ children }: { children: ReactNode }) {
   const location = useLocation();
   const { data: user } = useAuth();
 
+  useEffect(() => {
```
Martinsos, line 17: Any reason why you moved this up?

Ok, so you check if more than 5 minutes passed, and if so, you update the lastActiveTime -> I guess you don't want to do too many updates so that is the reason, and 5 minutes is good enough resolution.

Before you didn't await the result, but you also don't await it in the new code -> you do await in the `updateUserActivity`, but you don't await the call to `updateUserActivity`, so if I am correct, the effect should be the same -> you are not await for that update at the end. So if that was the issue, I would be surprised if this fixed it.

Can you confirm this fixes the issue, were you able to reproduce it?

So `updateCurrentUser` is an operation (action) that updates User. Meaning when it executes, due to how Operations work in Wasp, it will let Wasp know that User was updated, and so this component will restart, triggering this code again, but the check for 5 minutes is the one that should then make sure nothing happens again.

If this was an issue before, I would be mostly suspect of that check for time, for those 5 minutes. Becaues I see you changed some small pieces of logic there. Maybe that check was wrong before and now it is correct? Because as I said, I doubt that you did anything with `await` there, or that `await` was an issue.

Another question -> why does this trigger on `[user]`? So if `user` object changes, we update user's last active timestamp. What is the idea here? The idea is basically, if `user` is logged-in/known, then we wnat to update it, is that it? So basically "as soon as I get/see user, I want to update it's last active timestamp". Ok, that makes sense I guess.

vincanger: > Any reason why you moved this up?
Nope. DIdn't realize it, actually.

> Can you confirm this fixes the issue, were you able to reproduce it?
The issue is caused when you remove the User return type from the `updateCurrentUser` action and change it to void, and no longer return the User update:

```ts
export const updateCurrentUser: UpdateCurrentUser<Partial<User>, void> = async (user, context) => {
  console.log('updating current user')
  if (!context.user) {
    throw new HttpError(401);
  }

  context.entities.User.update({
    where: {
      id: context.user.id,
    },
    data: user,
  });
};
```
I see now that the correct solution in this case would be to await the`context.entities.User.update()` call within the operation. So you're right, this update does nothing for the current code.

Martinsos: Oh yes, you are right! Because if you returned it previosly, that means that you returned promise, and execution of the whole function will be considered done (if awaited, and it is, by the server), only once that promise resolves. But if you removed that return and didn't await, the operation would likely finish before it actually updates the user. So yes, putting `await` here is the important part!

aiddroid: 5 min is unreasonable

vincanger: In that case, I am going to close this since this is specific to that user's changes.

PR #255, opened 2024-08-22
Bumps Prisma to v5

infomiho:
Preparing Open SaaS for an upcoming Wasp version where Prisma 5 is used.


Martinsos (commented):
@infomiho LGTM, but I would also let @vincanger take a quick look.

PR #264, opened 2024-08-30
Replacing a tags with link tags in AppNavBar

### LINKED ISSUE #254: Make navigation in template use Link tags rather then A tags.
Currently, the template uses a tags for its navigation items.

Screenshot 2024-08-22 at 7 54 11 AM

https://github.com/wasp-lang/open-saas/blob/2a4c5e7206d2124d41138da9ce3e600c19f1b057/template/app/src/client/components/AppNavBar.tsx#L47-L54

This causes the page to reload when navigating around items. Example:

https://github.com/user-attachments/assets/76ba1e85-893c-487f-8312-87113ba4af27

My fix is quite a minor one. Replace the a tag used in the navigation mapping with the link tag from react-router-dom. Doing so creates a better nav experience for users (both mobile and nav will need to be updated). Example:

Screenshot 2024-08-22 at 8 06 29 AM

https://github.com/user-attachments/assets/960cf97d-705b-4e26-8fa9-c57e3b9aec00

Currently, Link is being used in this file from the wasp router, which I believe can only point to files defined in the main.wasp file. Are we able to replace the wasp version of Link with the react-router-dom, which will allow links to be defined and used within the nav file?

I would love to work and contribute on this. Its a small improvement, but having to reload pages in a react-app takes away from that seamless SPA experience.

npenza:
## Description

> Describe your PR! If it fixes specific issue, mention it with "Fixes # (issue)".
Fixes #254. Replaces anchor tags with link tags to prevent full page reloads when navigating the app navbar on both desktop and mobile.

I had to restructure the navigation array as outbound links can't use the link tags, and must use anchor tags. They are rendered after mapping through all the prior navigation items. Happy to discuss/implement a better solution.

## Contributor Checklist

### DIFF
--- a/template/app/src/client/components/AppNavBar.tsx
+++ b/template/app/src/client/components/AppNavBar.tsx
@@ -1,5 +1,6 @@
-import { Link, routes } from 'wasp/client/router';
+import { routes } from 'wasp/client/router';
 import { useAuth } from 'wasp/client/auth';
+import { Link } from 'react-router-dom';
 import { useState } from 'react';
 import { Dialog } from '@headlessui/react';
 import { BiLogIn } from 'react-icons/bi';
@@ -15,8 +16,6 @@ const navigation = [
   { name: 'AI Scheduler (Demo App)', href: routes.DemoAppRoute.build() },
   { name: 'File Upload (AWS S3)', href: routes.FileUploadRoute.build() },
   { name: 'Pricing', href: routes.PricingPageRoute.build() },
-  { name: 'Documentation', href: DocsUrl },
-  { name: 'Blog', href: BlogUrl },
 ];
 
 const NavLogo = () => <img className='h-8 w-8' src={logo} alt='Your SaaS App' />;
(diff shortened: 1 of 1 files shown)


vincanger:
Thanks for the contribution @npenza! I'm wondering why we can't use the `Link` component that Wasp exports since most (all?) of these links are routes defined in the `main.wasp` file and could benefit from its typesafety feature. Did you try it out and run into any issues?

npenza:
Thanks @vincanger. We definitely can use Wasp's `Link` component.

Before I make the changes, can you please clarify the differences between .build() and .to for the route objects? E.g this works:

```
const navigation = [
  { name: 'AI Scheduler (Demo App)', href: routes.DemoAppRoute.to },
  { name: 'File Upload (AWS S3)', href: routes.FileUploadRoute.to },
  { name: 'Pricing', href: routes.PricingPageRoute.to },
];
```

But this causes a ts error 'Type 'string' is not assignable to type' when  mapping through links

```
const navigation = [
  { name: 'AI Scheduler (Demo App)', href: routes.DemoAppRoute.build() },
  { name: 'File Upload (AWS S3)', href: routes.FileUploadRoute.build() },
  { name: 'Pricing', href: routes.PricingPageRoute.build() },
];

```

vincanger:
@npenza check out the proposed solution I gave above. The Wasp Link component expects a type which is a union of the route strings, so it functions more like an enum. That's why it won’t accept a general string, which is the error you referenced. The reason the build method throws this error is because it’s typed to return a string, which doesn’t match the union of route strings.

PR #267, opened 2024-09-09
Replace favicon

mmaksimovic:
## Description

Instead of using Wasp's logo as a favicon, use a free bee icon template.


mmaksimovic:
@sodic Fixed!

Looks like we have some flaky tests, the last commit I made was empty, and the tests passed 😅

sodic:
>  the last commit I made was empty, and the tests passed

No worries, that's perfectly fine. If tests are passing before an empty commit should, they should be passing on the empty commit too (since it didn't change anything).

What's strange is that one of the tests failed in the first place, but it seems to have been a network issue. I'll look into it right away.

Btw, there's still one `.DS_Store` left :)

mmaksimovic:
ahhh, my bad, done!

.gitignore
```diff
@@ -1,2 +1,3 @@
 /template/app/migrations
 /template/app/package-lock.json
+.DS_Store
```
infomiho, line 3: Looks good 👍 we have the same in the Wasp compiler `.gitignore`

PR #270, opened 2024-09-12
improve logic for no lemon squeezy customer portal url

vincanger:
we had a user confused by the current console.log statement `User with ID 123... does not have a LemonSqueezy customer portal URL` so I removed the log and added a note to the file. I also changed a bit of the logic.


template/app/src/payment/operations.ts
```diff
@@ -39,7 +38,7 @@ export const generateCheckoutSession: GenerateCheckoutSession<PaymentPlanId, Che
   };
 };
 
-export const getCustomerPortalUrl: GetCustomerPortalUrl<void, string | undefined> = async (_args, context) => {
+export const getCustomerPortalUrl: GetCustomerPortalUrl<void, string | null> = async (_args, context) => {
```
vincanger, line 41: I changed this to null since that's what prisma returns if the user hasn't had a customer portal url assigned to it yet by Lemon Squeezy

Martinsos (approved):

PR #273, opened 2024-09-16
Fix docs links & other small updates

vincanger:
A user in discord notified us of some dead links, so I went through and actually found many and updated them. Along with that, I made some small updates that I overlooked with the recent Lemon Squeezy PR


Martinsos (approved):
Nice!

PR #314, opened 2024-10-30
Use webp over png and lazy load images

vincanger:
## Description

Some minor image performance boosts

## Contributor Checklist


template/app/src/client/components/AppNavBar.tsx
```diff
@@ -19,7 +19,7 @@ const navigation = [
   { name: 'Blog', href: BlogUrl },
 ];
 
-const NavLogo = () => <img className='h-8 w-8' src={logo} alt='Your SaaS App' />;
+const NavLogo = () => <img className='h-8 w-8' src={logo} decoding='async' alt='Your SaaS App' />;
```
Martinsos, line 22: Interesting, article here claims there isn't much improvement normally https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decoding , which suggests one should normally not bother with setting this value.

How have you decided it is worth setting it?

template/app/src/landing-page/components/Testimonials.tsx
```diff
@@ -22,7 +22,7 @@ export default function Testimonials({ testimonials }: { testimonials: Testimoni
                 </blockquote>
                 <figcaption className='mt-6 text-base text-white'>
                   <a href={testimonial.socialUrl} className='flex items-center gap-x-2'>
-                    <img src={testimonial.avatarSrc} className='h-12 w-12 rounded-full' />
+                    <img src={testimonial.avatarSrc} loading='lazy' decoding='async' className='h-12 w-12 rounded-full' />
```
Martinsos, line 25: I did a quick read about `loading` here: https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading .

It says that in Firefox it has to go before `src` to take effect. Silly requirement, but maybe let's make sure that happens? Maybe React does it for us, but likely it doesn't, and better not to rely on it.

Also, I am guessing we shouldn't be setting `loading='lazy'` for every image we have? Or should we? What is the strategy about it, how do we make decision what is loaded lazily and what eagerly? I read that one con of lazy loading is that things around image might need to move around if height and width are not known upfront.

Martinsos (approved):
Approved, but ask some questions!

template/app/src/landing-page/components/Testimonials.tsx
vincanger: I thought that for everything below the main fold, lazy should be fine. I also tested it on the main hero image, which is kind of below the fold, and I don't see any noticeable difference.

template/app/src/client/components/AppNavBar.tsx
vincanger: Yeah, you're right, after reading up on it it seems like it's more likely designed for swapping out a thumbnail with a large image once it's done loading, something we're not doing here

template/app/src/landing-page/components/Testimonials.tsx
Martinsos: Ok soudns good!

template/app/src/client/components/AppNavBar.tsx
Martinsos: Ok! So we remove this one/

PR #318, opened 2024-11-15
Add proxied Plausible analytics

infomiho:
- This is a draft PR until we setup Plausible on plauisble.io
- It uses the proxied URL via Cloudflare Workers
- I deleted the `.local.js` second script since `local` is used if you want to track events on `localhost` - which we probably don't want
- I resolved a conflict in the `Header` component so I got this diff

### DIFF
--- a/opensaas-sh/app_diff/main.wasp.diff
+++ b/opensaas-sh/app_diff/main.wasp.diff
@@ -1,6 +1,6 @@
 --- template/app/main.wasp
 +++ opensaas-sh/app/main.wasp
-@@ -3,30 +3,30 @@
+@@ -3,30 +3,29 @@
      version: "^0.15.0"
    },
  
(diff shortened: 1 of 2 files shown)


opensaas-sh/app_diff/main.wasp.diff
```diff
@@ -43,12 +43,12 @@
...
 -    "<script defer data-domain='<your-site-id>' src='https://plausible.io/js/script.js'></script>",  // for production
 -    "<script defer data-domain='<your-site-id>' src='https://plausible.io/js/script.local.js'></script>",  // for development
-+    "<script defer data-domain='opensaas.sh' src='https://plausible.apps.twoducks.dev/js/script.js'></script>",
-+    "<script defer data-domain='opensaas.sh' src='https://plausible.apps.twoducks.dev/js/script.local.js'></script>",
++    "<script defer data-domain='opensaas.sh' data-api='/waspara/wasp/event' src='/waspara/wasp/script.js'></script>",
    ],
  
    // 🔐 Auth out of the box! https://wasp-lang.dev/docs/auth/overview
-@@ -38,7 +38,7 @@
+\ No newline at end of file
```
vincanger, line 50: what are all these `\ No newline...` additions in the main.wasp diff?

Martinsos: So I actually don't know, but I guess that in one of the files the newline at the end of the file got rmoved so that is why .diff shows that?

infomiho: I get like 5, 6 files with just `No newline at end of file` as changes, but I don't commit them. I'm not sure about what to do next? I just want to have this:
`<script defer data-domain='opensaas.sh' data-api='/waspara/wasp/event' src='/waspara/wasp/script.js'></script>` in the `main.wasp` and don't want to mess anything up.

PR #320, opened 2024-11-18
opensaas.sh stats improvements

infomiho:
- It updated the Plausible API call to include imported data (old stats will be included)
- It sorts the sources tables


opensaas-sh/app_diff/src/admin/dashboards/analytics/AnalyticsDashboardPage.tsx.diff
```diff
@@ -36,14 +41,24 @@
 +      console.error(error);
 +    }
 +  };
++
++  const sortedSources = useMemo(() => {
```
infomiho, line 45: Actual change, the rest is automatic formatting by Prettier.

vincanger (approved):
Noice

PR #337, opened 2024-12-04
Free vs paid boilerplates

mmaksimovic:
Added Matija as an author
Added images
Added a blog post


vincanger (changes requested):
Everything looks good except we're missing the `boilerplateLicenses.png` image

PR #356, opened 2025-01-23
Adds tailwindcss as explicit dep

infomiho:
Fixes this issue: https://github.com/wasp-lang/wasp/issues/2464

We should merge this ASAP and make it part of the **0.15.0 tag**


opensaas-sh/app_diff/src/analytics/stats.ts.diff
```diff
@@ -6,8 +6,9 @@
  import { stripe } from '../payment/stripe/stripeClient'
 -import { listOrders } from '@lemonsqueezy/lemonsqueezy.js';
  import { getDailyPageViews, getSources } from './providers/plausibleAnalyticsUtils';
- // import { getDailyPageViews, getSources } from './providers/googleAnalyticsUtils;
+-// import { getDailyPageViews, getSources } from './providers/googleAnalyticsUtils';
```
infomiho, line 9: There were some merge conflicts when I applied the patch, so I've accepted the version that doesn't have the 
```
import { paymentProcessor } from '../payment/paymentProcessor';
```
import.

Issue #208, opened 2024-07-02
forward https://saas-template-gpt-client.fly.dev/ to opensaas.sh


vincanger:
the old saas template demo was deployed to https://saas-template-gpt-client.fly.dev/

let's remove it and have it automatically redirect to opensaas.sh

PR #361, opened 2025-02-08
Add redirect from old demo URL to new URL 

pshycodr:
issue: #208 
---
Add redirection rule to `fly-client.toml` to redirect traffic from `https://saas-template-gpt-client.fly.dev/` to `https://opensaas.sh`.

* Add `[http_service.http_options.redirects]` section to handle redirection.
* Add rule to redirect all traffic from `https://saas-template-gpt-client.fly.dev/` to `https://opensaas.sh`.


vincanger:
one for you @infomiho

infomiho:
I'm not sure if editing the opensaas.sh client TOML file will add the redirect to the old client. We should probably edit the old demo client TOML file. @vincanger is it in the old saas template repo?

infomiho:
@pshycodr Yes, that would be the right course of action. I can see in the old template repo we don't have the TOML files for deployment, so you might not be able to do it. 

@vincanger do you have the TOML files locally or?

pshycodr:
@infomiho  @vincanger  Got it! If the TOML files aren’t in the old template repo, let me know where I can find them or if there’s another way to set up the redirect. Happy to proceed once I have the right files!

vincanger:
@infomiho I can't seem to find the old toml files. What could we do instead?

pshycodr:
Since we can't find the old `fly.toml` files, here are a few approaches we can take:

1. If Fly.io allows changes via the dashboard, we can try updating the settings there.
2. We can attempt to fetch the existing config with `flyctl config save` and update the redirect rules directly.
3. If we have access to the app's code, we can implement a server-side redirect.

vincanger:
I'm going to close this for now since a lot of issues are piling up and the old repo points to the new one anyway.

PR #362, opened 2025-02-09
no need to cast prisma return to User

Zeko369:


vincanger:
thanks! this was taken care of in #383 since we decided to remove the entire newsletter section anyway.

Yherealtita (commented):
Done

PR #367, opened 2025-02-17
Reviewing Open Saas docs

sodic:


opensaas-sh/blog/src/content/docs/start/getting-started.mdx
sodic, line 1: Open Saas docs look a little broken (notice the alignment, padding, banner in the middle of the docs, etc):

alt text

sodic (commented):
Superficial stuff I noticed while getting familiar with how Open Saas works

PR #372, opened 2025-02-18
Fix imports to avoid client-side errors

### LINKED ISSUE #370: organize `server/utils.ts` functions so that they aren't being imported on the client
We're importing `requireNodeEnvVar` into the `payment/plans.ts` and then some of these variables are being imported onto the client, causing clientside errors if the user extends the `utils.ts` file with a serverside only function.

See this convo for more info: https://discord.com/channels/686873244791210014/1339549500158771250/1339549500158771250

pshycodr:
Fixes #370

Organize `server/utils.ts` functions to avoid client-side imports.

* Add `template/app/src/server/envUtils.ts` to contain the `requireNodeEnvVar` function.
* Remove `requireNodeEnvVar` function from `template/app/src/server/utils.ts`.
* Update imports of `requireNodeEnvVar` in `template/app/src/payment/plans.ts`, `template/app/src/payment/lemonSqueezy/paymentProcessor.ts`, `template/app/src/payment/stripe/paymentProcessor.ts`, `template/app/src/payment/stripe/stripeClient.ts`, `template/app/src/payment/lemonSqueezy/webhook.ts`, and `template/app/src/payment/stripe/webhook.ts` to import from `template/app/src/server/envUtils.ts`.

### DIFF
--- a/template/app/src/payment/lemonSqueezy/paymentProcessor.ts
+++ b/template/app/src/payment/lemonSqueezy/paymentProcessor.ts
@@ -1,5 +1,5 @@
 import type { CreateCheckoutSessionArgs, FetchCustomerPortalUrlArgs, PaymentProcessor } from '../paymentProcessor';
-import { requireNodeEnvVar } from '../../server/utils';
+import { requireNodeEnvVar } from '../../server/envUtils';
 import { createLemonSqueezyCheckoutSession } from './checkoutUtils';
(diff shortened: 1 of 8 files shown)


infomiho:
Thank you for your effort! Right direction for this problem is reorganizing the code to prevent accidental imports of server code by transitive imports. We would love to accept a solution like that, where it is also obvious that careful thought was put into it. 

Right now, what you submitted seems more simplistic and likely LLM driven, without deeper understanding of the issue at hand.

PR #373, opened 2025-02-18
Add more detail about SKIP_EMAIL_VERIFICATION_IN_DEV to e2e tests README

### LINKED ISSUE #363: add more detail about SKIP_EMAIL_VERIFICATION_IN_DEV to e2e tests README
The signup process is not being verified when running e2e tests on a fresh Open SaaS instance.

- `SKIP_EMAIL_VERIFICATION_IN_DEV=true` needs to be added to the `app/.env.server` file to allow the tests to run. This is mentioned in the e2e tests README but could give more detailed instructions

pshycodr:
Fixes #363

Add `SKIP_EMAIL_VERIFICATION_IN_DEV=true` to `template/app/.env.server.example` file.

* Add a new section to the file to skip email verification in development mode.
* Add `SKIP_EMAIL_VERIFICATION_IN_DEV=true` to the new section.


vincanger:
Hey @pshycodr we actually would like to allow the user to decide for themselves if they want to skip email verification in development. Is there a reason you think it should be in there by default (if that reason is e2e tests, then you'll see we instruct the user to add it before running them in the e2e-tests README file).

pshycodr:
@vincanger 
The reason for having the SKIP_EMAIL_VERIFICATION_IN_DEV environment variable set to true by default in the e2e tests is to ensure that the tests run smoothly without manual intervention.

vincanger:
> @vincanger The reason for having the SKIP_EMAIL_VERIFICATION_IN_DEV environment variable set to true by default in the e2e tests is to ensure that the tests run smoothly without manual intervention.

but we have an explicit mention of it in the `e2e-tests/README.md`:

> Open another terminal and start the Wasp app with the environment variable set to skip email verification in development mode:
> ```shell
> cd app && SKIP_EMAIL_VERIFICATION_IN_DEV=true wasp start
> ```

If we add this to the app by default, then users may not understand that the auth method sends an email verification link by default. They also won't be able to test their customize verification email they'd like to send. I think it's best we leave it as is.

pshycodr:
Hey @vincanger,
Thanks for the clarification. You're  right, setting the variable by default might lead users to overlook email verification behavior during development.
I will update the PR to :
 - Remove the default `SKIP_EMAIL_VERIFICATION_IN_DEV=true` from `.env.server.example`.
 - Add clearer, explicit instructions in the e2e-tests/README.md on how to set this variable before running e2e tests. 
Let me know if you have any additional suggestions. And Thanks again for the feedback

pshycodr:
@vincanger 
I updated the `e2e-tests/README.md` file, Added more detailed explanations about the `SKIP_EMAIL_VERIFICATION_IN_DEV` variable

template/app/.env.server.example
```diff
@@ -53,4 +53,7 @@ GOOGLE_ANALYTICS_PROPERTY_ID=123456789
 AWS_S3_IAM_ACCESS_KEY=ACK...
 AWS_S3_IAM_SECRET_KEY=t+33a...
 AWS_S3_FILES_BUCKET=your-bucket-name
-AWS_S3_REGION=your-region
\ No newline at end of file
+AWS_S3_REGION=your-region
+
+# Skip email verification in development mode
```
vincanger, line 58: Let's remove this. We don't want this to be the default behaviour when developing, so that the users understand the verification flow.

template/e2e-tests/README.md
```diff
@@ -18,13 +18,28 @@ Start your Wasp DB and leave it running:
...
+> 
+> **However, during e2e tests, this manual step will cause the tests to hang and fail** because the link is never clicked. To prevent this, set the following environment variable when starting your app:
+
+```bash
 cd app && SKIP_EMAIL_VERIFICATION_IN_DEV=true wasp start
 ```
 
-> [!IMPORTANT]  
-> When using the email auth method a verification link is sent when the user registers, or logged to the console if you're using the default Dummy provider. You must click this link to complete registration. Setting SKIP_EMAIL_VERIFICATION_IN_DEV to "true" skips this verification step, allowing you to automatically log in. This step must be skipped when running tests, otherwise the tests will hang and fail as the verification link is never clicked!
+#### What this step will do:
```
vincanger, line 34: nice!

pshycodr: thanks ☺️

vincanger (approved):
Great. Thanks for the improvement!

pshycodr:
Glad to contribute. 🚀☺️

PR #375, opened 2025-02-19
improve one-time payment webhook logic

### LINKED ISSUE #371: add `payment_intent.succeeded` webhook handler to stripe webhook
At the moment, we favor subscription products, and our webhook handlers work fine for that. But we are not optimally processing one-time payment products at the moment.

For one-time payment products, with the Stripe mode `payments`, an invoice is not created. Therefore, instead of confirming payment in the `invoice.paid` or `checkout.session.completed` event handlers, we need to create a new event handler for the `payment_intent.succeeded` event. We do this because, although it seems rare, `checkout.session.completed` could fire even when a payment is unsuccessful.

To improve it we should:
- add a `payment_intent.succeeded` webhook handler and only confirm payment through this handler.
- remove the payment accepted logic (or perhaps the entire handler) for `checkout.session.completed`, as it can happen that checkout sessions complete but payments are still pending/fail.

vincanger:
## Description

Fixes #371 

Adds a new webhook event handler for `payment_intent.succeeded`. One-time payment products (i.e. "credits"), characterized by Stripe as having the `payment` mode (not the `subscription` mode), do not send an `invoice.paid` event (although subscriptions do), so we have to confirm payment and handle it better than just using the `checkout.session.completed` event, which can be sent even in the case that a payment fails.

## Contributor Checklist

### DIFF
--- a/template/app/src/payment/stripe/webhook.ts
+++ b/template/app/src/payment/stripe/webhook.ts
@@ -33,6 +33,11 @@ export const stripeWebhook: PaymentsWebhook = async (request, response, context)
       const invoice = event.data.object as Stripe.Invoice;
       await handleInvoicePaid(invoice, prismaUserDelegate);
       break;
+    case 'payment_intent.succeeded':
+      const paymentIntent = event.data.object as Stripe.PaymentIntent;
+      if (paymentIntent.invoice) return; // We handle invoices in the invoice.paid webhook.
+      await handlePaymentIntentSucceeded(paymentIntent, prismaUserDelegate);
+      break;
     case 'customer.subscription.updated':
       const updatedSubscription = event.data.object as Stripe.Subscription;
       await handleCustomerSubscriptionUpdated(updatedSubscription, prismaUserDelegate);
@@ -59,6 +64,9 @@ export const stripeMiddlewareConfigFn: MiddlewareConfigFn = (middlewareConfig) =
   return middlewareConfig;
 };
 
+// Because a checkout session completed could potentially result in a failed payment,
+// we can update the user's payment details here, but confirm credits or a subscription
+// if the payment succeeds in other, more specific, webhooks.
 export async function handleCheckoutSessionCompleted(
   session: Stripe.Checkout.Session,
   prismaUserDelegate: PrismaClient["user"]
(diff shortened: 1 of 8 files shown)


template/app/src/payment/stripe/webhook.ts
```diff
@@ -174,4 +193,23 @@ function getPlanIdByPriceId(priceId: string): PaymentPlanId {
...
+      break;
+    case 'credits':
+      numOfCreditsPurchased = plan.effect.amount;
+      break;
+    default:
+      assertUnreachable(plan.effect);
+  }
+
+  return { subscriptionPlan, numOfCreditsPurchased };
+}
```
infomiho, line 215: Effective Typescript suggests constructing whole objects, so I'd go with something like this:
```ts
function getPlanEffectDetailsById(planId: PaymentPlanId): {
  subscriptionPlan: PaymentPlanId | undefined;
  numOfCreditsPurchased: number | undefined;
} {
  const plan = paymentPlans[planId];

  switch (plan.effect.kind) {
    case 'subscription':
      return { subscriptionPlan: planId, numOfCreditsPurchased: undefined };
    case 'credits':
      return { subscriptionPlan: undefined, numOfCreditsPurchased: plan.effect.amount };
    default:
      assertUnreachable(plan.effect);
  }
}
```

template/app/src/payment/stripe/webhook.ts
```diff
@@ -67,37 +75,48 @@ export async function handleCheckoutSessionCompleted(
...
-    case 'subscription':
-      subscriptionPlan = planId;
-      break;
-    case 'credits':
-      numOfCreditsPurchased = plan.effect.amount;
-      break;
-    default:
-      assertUnreachable(plan.effect);
-  }
+  const { subscriptionPlan } = getPlanEffectDetailsById(planId);
```
infomiho, line 80: Here it feels like we are implicitly saying: we only want to update the `subscriptionPlan` value for `subscription` plans by not using the `numOfCreditsPurchased` value here. Can we be explicit about it? Something like:

```ts
if (plan.effect.kind === 'credits') {
  return;
}
```

template/app/src/payment/stripe/webhook.ts
```diff
@@ -67,37 +75,48 @@ export async function handleCheckoutSessionCompleted(
...
+  const { metadata } = paymentIntent;
+
+  if (!metadata.priceId) {
+    throw new HttpError(400, 'No price id found in payment intent');
+  }
+
+  const planId = getPlanIdByPriceId(metadata.priceId);
+  const { subscriptionPlan, numOfCreditsPurchased } = getPlanEffectDetailsById(planId);
+
+  return updateUserStripePaymentDetails(
```
infomiho, line 114: I feel like the `updateUserStripePaymentDetails` method is doing too much, it's used for:
- updating `subscriptionPlan`
- updating `datePaid`
- incrementing the credits
- updating `subscriptionStatus`

I'd recommend that we refactor it into multiple smaller methods that could be very specific about what they are updating:
- `updateSubscriptionPlan`
- `updateCreditsAmount`
- `updateDatePaid`
etc.

There are way to make this composable, we compose the `data` object for multiple updates e.g.
```ts
updateUserStripePaymentDetails([
  createDatePaidUpdate(datePaid),
  createSubscriptionStatusUpdate(subscriptionStatus),
])
```

I'm not saying to do this in this PR, but rather in a separate PR that only refactors this part.

infomiho (commented):
If I understood correctly, we want to give users access when we are sure that payment succeeded. That's why we introduced a new hook `payment_intent.succeeded`.

I've left some comments.

template/app/src/payment/stripe/webhook.ts
vincanger: Good idea. I created this issue https://github.com/wasp-lang/open-saas/issues/376

infomiho (approved):
Good job 👍

PR #379, opened 2025-02-20
Adds runtime validation for operation inputs

infomiho:
- I've added Zod schemas for all operation inputs
- There is a new `ensureArgsSchemaOrThrowHttpError` helper that I used in all the operations
- I've updated the S3 upload validation to execute on server (as well as on the client)


template/app/src/demo-ai-app/operations.ts
```diff
@@ -20,15 +22,23 @@ function setupOpenAI() {
...
 
-export const generateGptResponse: GenerateGptResponse<GptPayload, GeneratedSchedule> = async ({ hours }, context) => {
+const generateGptResponseInputSchema = z.object({
+  hours: z.string().regex(/^\d+(\.\d+)?$/, 'Hours must be a number'),
+});
+
+type GenerateGptResponseInput = z.infer<typeof generateGptResponseInputSchema>;
+
+export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput, GeneratedSchedule> = async (
+  rawArgs: unknown,
```
infomiho, line 33: This is one important detail, this changes the type of `rawArgs` to `unknown` _until it's validated_ so people don't use `rawArgs` and think it's of some shape - when it's not.

This still works with our full stack type safety @sodic since the system will take the `GenerateGptResponse<GenerateGptResponseInput, GeneratedSchedule>` type and not bother with reading the actual `rawArgs` type.

template/app/src/demo-ai-app/operations.ts
```diff
@@ -20,15 +22,23 @@ function setupOpenAI() {
...
+  context
+) => {
   if (!context.user) {
     throw new HttpError(401);
   }
 
+  const args = ensureArgsSchemaOrThrowHttpError(generateGptResponseInputSchema, rawArgs);
```
infomiho, line 40: Take the raw args, validate them and return them as properly typed `args` which can be then safely used in the rest of the operation.

template/app/src/demo-ai-app/operations.ts
```diff
@@ -20,15 +22,23 @@ function setupOpenAI() {
...
-type GptPayload = {
-  hours: string;
-};
 
-export const generateGptResponse: GenerateGptResponse<GptPayload, GeneratedSchedule> = async ({ hours }, context) => {
+const generateGptResponseInputSchema = z.object({
+  hours: z.string().regex(/^\d+(\.\d+)?$/, 'Hours must be a number'),
+});
+
+type GenerateGptResponseInput = z.infer<typeof generateGptResponseInputSchema>;
```
infomiho, line 30: The type that will be used with the operation type is inferred from the schema - so the runtime and type info is in sync!

template/app/src/demo-ai-app/operations.ts
```diff
@@ -20,15 +22,23 @@ function setupOpenAI() {
 }
 
 //#region Actions
-type GptPayload = {
-  hours: string;
-};
 
-export const generateGptResponse: GenerateGptResponse<GptPayload, GeneratedSchedule> = async ({ hours }, context) => {
+const generateGptResponseInputSchema = z.object({
```
infomiho, line 26: Define a schema for the input.

template/app/src/file-upload/fileUploading.ts
```diff
@@ -37,18 +23,48 @@ export async function uploadFileWithProgress({ file, setUploadProgressPercent }:
...
+type FileWithValidType = Omit<File, 'type'> & { type: AllowedFileType };
+
+type FileParseResult =
+  | { kind: 'success'; file: FileWithValidType }
+  | {
+      kind: 'error';
+      error: { message: string; code: 'INVALID_FILE_TYPE' | 'FILE_TOO_LARGE' };
+    };
+
+export function parseValidFile(file: File): FileParseResult {
```
infomiho, line 41: I went with parsing (we parse a raw file into a file which has the `type` only of certain values) so Typescript wouldn't complain when I use the `file.type` in the `createFile` action.

template/app/src/file-upload/s3Utils.ts
```diff
@@ -13,27 +14,30 @@ const s3Client = new S3Client({
...
-}
+  fileName: string;
+  userId: string;
+};
 
-export const getUploadFileSignedURLFromS3 = async ({fileType, userInfo}: S3Upload) => {
-  const ex = fileType.split('/')[1];
-  const Key = `${userInfo}/${randomUUID()}.${ex}`;
-  const s3Params = {
+export const getUploadFileSignedURLFromS3 = async ({ fileName, fileType, userId }: S3Upload) => {
```
infomiho, line 21: Added `fileName` here so I can extract the extension properly.

template/app/src/file-upload/s3Utils.ts
```diff
@@ -13,27 +14,30 @@ const s3Client = new S3Client({
...
     Bucket: process.env.AWS_S3_FILES_BUCKET,
     Key: key,
-  };
-  const command = new GetObjectCommand(s3Params);
+  });
   return await getSignedUrl(s3Client, command, { expiresIn: 3600 });
+};
+
+function getS3Key(fileName: string, userId: string) {
+  const ext = path.extname(fileName).slice(1);
```
infomiho, line 41: Extracted the extension from the `fileName` and not from MIME type as done previously. This is not related to the runtime validations, but it was an small enough change.

template/app/src/server/validation.ts
```diff
@@ -0,0 +1,14 @@
...
+import * as z from 'zod';
+
+export function ensureArgsSchemaOrThrowHttpError<Schema extends z.ZodType<any, any>>(
+  schema: Schema,
+  rawArgs: unknown
+): z.infer<Schema> {
+  const parseResult = schema.safeParse(rawArgs);
+  if (!parseResult.success) {
+    console.error(parseResult.error);
+    throw new HttpError(400, 'Operation arguments validation failed', { errors: parseResult.error.errors });
```
infomiho, line 11: Each time the raw args validation fails, throw an 400 error and return the validation details in the error data object. This can be used on the client to show a more detailed message e.g. `The validation failed on the "id" field`

template/app/src/server/validation.ts
```diff
@@ -0,0 +1,14 @@
+import { HttpError } from 'wasp/server';
+import * as z from 'zod';
+
+export function ensureArgsSchemaOrThrowHttpError<Schema extends z.ZodType<any, any>>(
```
infomiho, line 4: Helper fn that either returns the properly typed args or throws an `HttpError`.

template/app/src/demo-ai-app/operations.ts
sodic: I understand it, but it's kind of specific (advanced). Should we document it somehow @vincanger?

template/app/src/server/validation.ts
sodic: Could we change the `any`s `z.ZodType` to something else?

template/app/src/demo-ai-app/operations.ts
```diff
@@ -181,47 +193,73 @@ export const generateGptResponse: GenerateGptResponse<GptPayload, GeneratedSched
...
+});
+
+type CreateTaskInput = z.infer<typeof createTaskInputSchema>;
+
+export const createTask: CreateTask<CreateTaskInput, Task> = async (rawArgs: unknown, context) => {
   if (!context.user) {
     throw new HttpError(401);
   }
 
+  const args = ensureArgsSchemaOrThrowHttpError(createTaskInputSchema, rawArgs);
```
sodic, line 207: Why not destructure it right away? It minimize the changes in the rest of the body and is more ergonomic. Same goes for other operations.

template/app/src/demo-ai-app/operations.ts
```diff
@@ -181,47 +193,73 @@ export const generateGptResponse: GenerateGptResponse<GptPayload, GeneratedSched
...
+});
+
+type UpdateTaskInput = z.infer<typeof updateTaskInputSchema>;
+
+export const updateTask: UpdateTask<UpdateTaskInput, Task> = async (rawArgs: unknown, context) => {
   if (!context.user) {
     throw new HttpError(401);
   }
 
+  const args = ensureArgsSchemaOrThrowHttpError(updateTaskInputSchema, rawArgs);
```
sodic, line 232: This one could also be destructured immediately.

template/app/src/file-upload/fileUploading.ts
```diff
@@ -1,30 +1,18 @@
 import { Dispatch, SetStateAction } from 'react';
 import { createFile } from 'wasp/client/operations';
 import axios from 'axios';
+import { ALLOWED_FILE_TYPES, MAX_FILE_SIZE } from './validation';
 
+type AllowedFileType = (typeof ALLOWED_FILE_TYPES)[number];
+type FileWithValidType = Omit<File, 'type'> & { type: AllowedFileType };
```
sodic, line 7: I can't understand where this type `File` is coming from.

template/app/src/file-upload/FileUploadPage.tsx
```diff
@@ -64,13 +65,12 @@ export default function FileUploadPage() {
...
       }
 
-      const validationError = validateFile(file);
-      if (validationError) {
-        setUploadError(validationError);
+      const validFileResult = parseValidFile(file);
+      if (validFileResult.kind === 'error') {
+        setUploadError(validFileResult.error);
         return;
       }
```
sodic, line 72: Hmm, again, I know what's going on and it's definitely better, but do you think it's too much for an average user?

Especially with the `Result` type and all.

template/app/src/server/validation.ts
sodic: Very minor, but I'd use structured programming here:
```typescript
if (!parseResult.success) {
  // ...
} else {
  return parseResult.data
}
```
It makes the control flow very explicit, which helps since we use `throw`. If someone removes the throw, we'll get a type error, which is nice.

template/app/src/user/operations.ts
sodic, line 1: We'll have some conflicts here, but nothing major.

Nice work!

template/app/src/server/validation.ts
sodic: Also, how does the full error response look like? Is it nice?

sodic (approved):
Great stuff, left some comments.

I have some reservations about advanced constructs you use, @vincanger let us know what you think, is it too much.

template/app/src/demo-ai-app/operations.ts
infomiho: Ideally, we'd upgrade Wasp to allow you to define operations with the Zod schema and this dance would be hidden from the user. Something like: 
```ts
defineAction(inputSchema, (args, context) => {})
```

template/app/src/file-upload/fileUploading.ts
infomiho: I believe it's built-in like `Map`

template/app/src/demo-ai-app/operations.ts
vincanger: Ooof yeah this is quite advanced. I think many of our uses won't understand what is going on with the `rawArgs: unknown`. In fact, I don't quite totally understand it.

vincanger: Ok after looking it over I get what's going on, but I have some questions regarding the `rawArgs: unknown1 typing:
1. Haven't we already typed rawArgs with `GenerateGptResponseInput`? Why are we re-typing it as unknown then?
2. I dropped the `unknown` from the `rawArgs` argument and saw no errors. We're still validating the type at runtime with your new `ensureArgsSchemaOrThrowHttpError()` function, so couldn't we just drope the `unknown` type and avoid confusion? I think this will turn some heads amongst the less experienced devs like myself.

vincanger (approved):
Looks good I just left some comments regarding complexity (for our users)

template/app/src/demo-ai-app/operations.ts
infomiho: > Haven't we already typed rawArgs with GenerateGptResponseInput? Why are we re-typing it as unknown then?

Yes, we did and that's part of the problem 😄 Users receive the `rawRags` typed without validation, so they think "Oh great, this is a `string`" but it could be `undefined`. Types can lie, run time validation can't. That's why I wanted to go the extra mile and make sure they _need to validate_ the args to get the types again. 

I couldn't write `unknown` in the `GenerateGptResponseInput` type because our full-stack type safety would no longer work.

Okay, since we want to tackle this at Wasp level, **I'll drop the `unknown`s for now to keep things cleaner.** Once we implement runtime validations API for Wasp operations - this will be 💯

template/app/src/server/validation.ts
infomiho: It looks like this:
```json
{
  "message": "Operation arguments validation failed",
  "data": {
    "errors": [
      {
        "code": "too_small",
        "minimum": 1,
        "type": "string",
        "inclusive": true,
        "exact": false,
        "message": "String must contain at least 1 character(s)",
        "path": [
          "description"
        ]
      }
    ]
  }
}
```

It's in a standard format which users can use to generate pretty error messages and/or show them under appropriate fields. It's quite generic IMHO as this error is really not expected to happen if they use client side validations with e.g. https://react-hook-form.com which is the standard way to do it.

template/app/src/server/validation.ts
infomiho: Actually, we don't need any `any`s here, we can just write `z.ZodTypeAny`.

This is what Zod docs recommend:
```ts
function parseData<T extends z.ZodTypeAny>(data: unknown, schema: T) {
  return schema.parse(data) as z.infer<T>;
  //                        ^^^^^^^^^^^^^^ <- add this
}

parseData("sup", z.string());
// => string
```

template/app/src/file-upload/FileUploadPage.tsx
infomiho: I'll simplify it by simply using type assertions 👍

template/app/src/file-upload/validation.ts
```diff
@@ -0,0 +1,9 @@
+export const MAX_FILE_SIZE = 5 * 1024 * 1024; // Set this to the max file size you want to allow (currently 5MB).
+export const ALLOWED_FILE_TYPES = [
```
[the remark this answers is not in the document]
sodic: I meant a newline, sorry. Not important anyway.

template/app/src/server/validation.ts
sodic: I like the info but this is a lot:
image

Can we hide it somehow?

template/app/src/user/operations.ts
```diff
@@ -30,41 +40,51 @@ export const updateIsUserAdminById: UpdateIsUserAdminById<{ id: string; data: Pi
...
-};
 type GetPaginatedUsersOutput = {
   users: Pick<User, 'id' | 'email' | 'username' | 'subscriptionStatus' | 'paymentProcessorUserId'>[];
   totalPages: number;
 };
 
+const getPaginatorArgsSchema = z.object({
+  skip: z.number(),
+  cursor: z.number().optional(),
+  emailContains: z.string().nonempty().optional(),
```
sodic, line 51: This caused en error on the admin dashboard - after you'd delete the email fitlter, the app would crash.

PR #384, opened 2025-02-25
Tighten up S3 CORS configuration

sodic:
TODO:
 - [x] Add reminder to set the domain in the deploy docs

## File uploading docs
image
## Deployment docs
image

@vincanger Can you please update the CORS config for opensaas.sh on AWS and test it out.

PR #385, opened 2025-02-25
Update S3 upload to validate file size and use POST

infomiho:
It uses the `createPresignedPost` to create an upload URL which can have limit the file upload file. It sets the `content-length-range` to be of maximum `MAX_FILE_SIZE_BYTES` which is the same value we use on the client.


sodic (approved):
Looks good to me!
I'm assuming you tested it (I didn't).

infomiho:
I've tested it 👍 

We should update the same docs as you did, after we merge #384

vincanger (approved):

opensaas-sh/blog/src/content/docs/guides/file-uploading.mdx
```diff
@@ -105,8 +105,6 @@ Now we need to change some permissions on the bucket to allow for file uploads f
        }
    ]
    ```
-   As an example, here are the CORS permissions for this site - https://opensaas.sh:
-   <Image src={corsExample} alt="cors-example" loading="lazy" />
```
infomiho, line 109: @sodic I've deleted this image since it's just repeating what the code block says and it's easier so we don't have to retake screenshots every time we update the code block

PR #390, opened 2025-02-26
Handle missing Discord email sooner. Make sure emails are verified.

infomiho:
This PR:
- Prevents user to sign up with Discord if they don't have an `email`
- Makes sure that the provided OAuth email is verified before used to set the `isAdmin` field
- Shows error message on the pricing page if something goes wrong

Screenshot 2025-02-26 at 15 29 26


opensaas-sh/app_diff/src/auth/userSignupFields.ts.diff
```diff
@@ -1,46 +1,64 @@
...
+   email: z.string(),
+ });
+@@ -16,10 +14,6 @@
+     const emailData = emailDataSchema.parse(data);
+     return emailData.email;
+   },
+-  isAdmin: (data) => {
+-    const emailData = emailDataSchema.parse(data);
+-    return adminEmails.includes(emailData.email);
+-  },
```
sodic, line 19: How come the example app removes this `isAdmin` check.

I can see it also does this in the old version of the code, but why?

sodic (approved):
Looks good.

I left just one question out of curiosity

template/app/src/payment/operations.ts
```diff
@@ -25,10 +25,8 @@ export const generateCheckoutSession: GenerateCheckoutSession<
   const userId = context.user.id;
   const userEmail = context.user.email;
   if (!userEmail) {
-    throw new HttpError(
-      403,
-      'User needs an email to make a payment. If using the usernameAndPassword Auth method, switch to an Auth method that provides an email.'
-    );
+    // If using the usernameAndPassword Auth method, switch to an Auth method that provides an email.
+    throw new HttpError(403, 'User needs an email to make a payment.');
```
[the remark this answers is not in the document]
infomiho: Yes, it didn't feel correct to give developer instructions in a user-facing error.

opensaas-sh/app_diff/src/auth/userSignupFields.ts.diff
infomiho: If you visit https://opensaas.sh/admin you'll notice you can see it with any account.

sodic: Ah, I had a feeling it has something to do with that.

But does that mean that the default behavior is that "if "adminicity" isn't defined, everyone is an admin?"

> you'll notice you can see it with any account.

For this feature, I'd expect `isAdmin` to be a function that just returns `true`.

infomiho: The default is set to `true` in the `opensaas.sh`

PR #391, opened 2025-02-28
Add database transactions where necessary

sodic:


template/app/src/analytics/operations.ts
```diff
@@ -12,31 +12,32 @@ type DailyStatsValues = {
...
+
+  const statsQuery = {
     orderBy: {
       date: 'desc',
     },
     include: {
       sources: true,
     },
-  });
+  } as const;
```
sodic, line 30: FYI: Necessary to prevent type widening.

For example, fi I didn't use `as const`, TS would infer the type of `statsQuery.orderBy.date` as `string`, which is to permissive to be accepted by `findFirst` and `findMany`.

template/app/src/demo-ai-app/operations.ts
```diff
@@ -35,11 +35,15 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
   context
 ) => {
```
sodic, line 36: This operation went through many changes.

It's not 100% foolproof (i.e., doesn't cover all the edge cases).

It can be done, but I can't do it in a way that's simple enough to make sense for an OpenSaas example app. I believe I'd need an extra table with pending responses and a job that periodically deletes unfinished responses and refunds the user.

In any case, I'm not merging this before a call with Vinny since I don't fully understand the old code. I left a bunch of `TODO`s and `NOTE`s with my questions.

template/app/src/demo-ai-app/operations.ts
```diff
@@ -48,150 +52,61 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
     },
   });
 
-  const parsedTasks = tasks.map(({ description, time }) => ({
-    description,
-    time,
-  }));
+  console.log('Calling open AI api');
```
infomiho, line 55: Leftover log

template/app/src/demo-ai-app/operations.ts
```diff
@@ -296,3 +211,98 @@ export const getAllTasksByUser: GetAllTasksByUser<void, Task[]> = async (_args,
   });
 };
 //#endregion
+
+// TODO: Why is hours a string?
```
infomiho, line 215: It can be easily be changed, update the operations validation, parse the `input` as number and change it here as well. I guess it was easier to just send the `string` that was received from `input` directly.

template/app/src/demo-ai-app/operations.ts
```diff
@@ -296,3 +211,98 @@ export const getAllTasksByUser: GetAllTasksByUser<void, Task[]> = async (_args,
...
+async function getDailyPlanFromGpt(tasks: Task[], hours: string): Promise<string | null> {
+  // TODO: Why was this a singleton
```
infomiho, line 217: It's like creating an instance of `axios` and using it everywhere. We init the OpenAI client and use it in multiple places. In some projects we would even put it in `openai.ts` and import it as a service we use. I'd keep it that way, since most users are used to that kind of usage.

template/app/src/demo-ai-app/operations.ts
sodic: Actually, I included this one on purpose :smile: 

We already had "Decrementing credits," so it seemed fitting, especially since the "decrementing credits" message now appears very late (after we get the response from GPT).

template/app/src/demo-ai-app/operations.ts
```diff
@@ -48,150 +52,61 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
...
-  }
+  // TODO: Can this ever fail?
+  return JSON.parse(dailyPlanJson);
 };
 
+function isEligibleForResponse(user: User) {
+  // TODO: Why not check for allowed states?
+  const isUserSubscribed =
+    user.subscriptionStatus !== SubscriptionStatus.Deleted &&
+    user.subscriptionStatus !== SubscriptionStatus.PastDue;
```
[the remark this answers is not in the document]
sodic: The best practice is usually to always default to the less damaging/more secure state. This pattern is called "fail-safe defaults" or "implicit deny."

It's the principle behind defining firewall rules (e.g., deny everything by default, specify a list of IPs and protocols you allow) and authorization (unauthorized unless proven otherwise).

The idea is that, if a new thing appears and you forget to update the rules, you want it to fail in a way that's safe. Let's assume you add a new state (allowed or unapproved) and forget to update this check.
It's better to temporarily reject some eligible users (they'll complain and you'll fix it quickly) than to allow non-eligible users (bad actors will drain your resoruces and keep it to themselves).

template/app/src/demo-ai-app/operations.ts
sodic: I had that idea too. This is called an interactive transaction (I think).

The thing is, transactions should be very short-lived. Calling an API over the network in a transaction is a dangerous (if the call hangs, it can block the entire database).

Prisma[ recommends against this](https://www.prisma.io/docs/orm/prisma-client/queries/transactions#overview) as well:
> Use interactive transactions with caution. Keeping transactions open for a long time hurts database performance and can even cause deadlocks. **Try to avoid performing network requests and executing slow queries inside your transaction functions**. We recommend you get in and out as quick as possible!

template/app/src/analytics/operations.ts
sodic: Hm, what exactly do you mean by typing it directly?

template/app/src/demo-ai-app/operations.ts
```diff
@@ -48,150 +52,61 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
...
-        },
-      });
-    }
-    console.error(error);
-    const statusCode = error.statusCode || 500;
-    const errorMessage = error.message || 'Internal server error';
-    throw new HttpError(statusCode, errorMessage);
-  }
+  // TODO: Can this ever fail?
+  return JSON.parse(dailyPlanJson);
```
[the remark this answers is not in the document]
sodic: But can GPT even return invalid JSON or is it deterministic?

Also, if it can return invalid json and this does fail, what do we want to do about user credits?

template/app/src/demo-ai-app/operations.ts
```diff
@@ -48,150 +52,61 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
...
+  // TODO: Why not check for allowed states?
+  const isUserSubscribed =
+    user.subscriptionStatus !== SubscriptionStatus.Deleted &&
+    user.subscriptionStatus !== SubscriptionStatus.PastDue;
+  const userHasCredits = user.credits > 0;
+  // TODO: If the user is subscribed (flat rate, use their subscription) -
+  // shouldn't take priority over credits since it makes no sece spending
+  // credits when a subscription is active?
+  // If they aren't subscribed, then it would make sense to spend credits.
+  // The old code always subtracted credits so I kept that, is this a bug?
```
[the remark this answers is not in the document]
sodic: Made an issue: https://github.com/wasp-lang/open-saas/issues/418

template/app/src/demo-ai-app/operations.ts
```diff
@@ -48,150 +52,61 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
...
-        },
-      },
-      temperature: 1,
-    });
+  // TODO: Do I need a try catch now that I'm saving a response and
+  // decrementing credits in a transaction?
 
-    const gptArgs = completion?.choices[0]?.message?.tool_calls?.[0]?.function.arguments;
+  // NOTE: I changed this up, first I do the request, and then I decrement
+  // credits. Is that dangerous? Protecting from it could be too complicated
```
sodic, line 65: @infomiho Please comment on this

template/app/src/demo-ai-app/operations.ts
```diff
@@ -48,150 +52,61 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
...
+  // TODO: Potential issue here, user can lose credits between the point we
+  // inject it into the context and here
```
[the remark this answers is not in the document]
sodic: > I don't understand what you mean here by "user can lose credits between the point we inject it into context and here".

Made an issue that explains it in detail (it's a core Wasp issue more than an Open Saas one): https://github.com/wasp-lang/wasp/issues/2675

> When using prisma transactions we have to run them at the end after the openai request, and I think that's fine.

This is a different problem, but I think it's fine too.

template/app/src/demo-ai-app/operations.ts
```diff
@@ -8,18 +8,19 @@ import type {
...
-    return new HttpError(500, 'OpenAI API key is not set');
+const openAi = getOpenAi();
+function getOpenAi(): OpenAI | null {
+  if (process.env.OPENAI_API_KEY) {
+    return new OpenAI({ apiKey: process.env.OPENAI_API_KEY });
+  } else {
+    return null;
   }
-  return new OpenAI({ apiKey: process.env.OPENAI_API_KEY });
 }
```
infomiho, line 24: This kind of change goes against what we have been doing in the framework ourselves: throw if env vars are missing immediately. We should migrate all of the env vars validation to use the new Zod env vars validation feature (in a separate PR of course) and when we do it - it will throw an error when the env vars is missing. To keep the behavior consistent - I'd do the same here.

template/app/src/demo-ai-app/operations.ts
infomiho: We are concerned here with the case **call to OpenAI failed**:
- if we **decrement before** -> user loses credits
- if we **decrement after** -> use can potentially spam many requests at once and abuse the system

I guess it depends on what the Open Saas user optimises for, for some it might be easier just to give users more credits if something goes wrong. Some maybe want to make sure users don't feel cheated. 

What I'm trying to say, I'm not sure which way to go - decrement before or after. I think it maybe makes sense to lay out the pros and cons in a comment so users can maybe adjust it to their needs?

template/app/src/demo-ai-app/operations.ts
infomiho: We are concerned here with the case **call to OpenAI failed**:
- if we **decrement before** -> user loses credits
- if we **decrement after** -> use can potentially spam many requests at once and abuse the system

I guess it depends on what the Open Saas user optimises for, for some it might be easier just to give users more credits if something goes wrong. Some maybe want to make sure users don't feel cheated. 

What I'm trying to say, I'm not sure which way to go - decrement before or after. I think it maybe makes sense to lay out the pros and cons in a comment so users can maybe adjust it to their needs?

template/app/src/demo-ai-app/operations.ts
```diff
@@ -48,150 +53,48 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
...
-      ],
-      tool_choice: {
-        type: 'function',
-        function: {
-          name: 'parseTodaysSchedule',
-        },
-      },
-      temperature: 1,
-    });
+  // TODO: Do I need a try catch now that I'm saving a response and
```
infomiho, line 62: I think it's better to throw a custom `HttpError` with some informative text than not. I think if you don't catch it, it will return an 500 with no text. That's also, okay, you decide.

template/app/src/demo-ai-app/operations.ts
infomiho: I think having credits as a number is the problem here, credits should somehow be calculated based on some list of usage e.g. there are 10 entries for successful GPT generations, that -10 credits. 

Another thing that could help us here is some sort of a queue of pending generations that "freeze" credits e.g. 2 pending generations temporarily block 2 credits and on success, decrement 2 credits. 

Maybe it's too complex for an example app, but maybe not, it's just one extra table + maybe a cleanup job. It maybe makes sense to create an issue for an "ideal system" if we want to go that way and solve it in the future.

infomiho (commented):
Left some comments and thoughts :)

vincanger (approved):
Noice 👊

template/app/src/demo-ai-app/operations.ts
sodic: Hm, not sure I understand. What would this text say? Which error are we talking about?

template/app/src/demo-ai-app/operations.ts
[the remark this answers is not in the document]
sodic: > I think that it can't return invalid JSON because we are using function calling with a certain JSON schema.

That's how it looks to me as well. @vincanger what makes you think the json string can be invalid?

template/app/src/demo-ai-app/operations.ts
sodic: We can revisit it, sure. I'll create an issue.

I still think an extra table and cleanup jobs are a bit too much though, but could be convinced otherwise :)

template/app/src/demo-ai-app/operations.ts
```diff
@@ -34,11 +35,14 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
...
-    throw new HttpError(401);
+    throw new HttpError(401, 'Only authenticated users are allowed to perform this operation');
   }
 
-  const { hours } = ensureArgsSchemaOrThrowHttpError(generateGptResponseInputSchema, rawArgs);
+  if (!isEligibleForResponse(context.user)) {
+    throw new HttpError(402, 'User has not paid or is out of credits');
+  }
 
+  const { hours } = ensureArgsSchemaOrThrowHttpError(generateGptResponseInputSchema, rawArgs);
```
sodic, line 45: We parse user input only after the user successfully authenticated.

template/app/src/demo-ai-app/operations.ts
```diff
@@ -48,150 +52,61 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
...
-    });
+  // TODO: Do I need a try catch now that I'm saving a response and
+  // decrementing credits in a transaction?
 
-    const gptArgs = completion?.choices[0]?.message?.tool_calls?.[0]?.function.arguments;
+  // NOTE: I changed this up, first I do the request, and then I decrement
+  // credits. Is that dangerous? Protecting from it could be too complicated
+  // TODO: Potential issue here, user can lose credits between the point we
+  // inject it into the context and here
+  // Seems to me that there should be users in the database with negative credits
```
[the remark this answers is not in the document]
sodic: Nice trick, but doesn't help us unfortunately because we still call the API between checking the user's eligibility and decrementing their credits.

I've decided it's no big deal after all and left a comment in the code.

infomiho (approved):

template/app/src/demo-ai-app/operations.ts
sodic: Done: https://github.com/wasp-lang/open-saas/issues/421

vincanger (approved):
i like

template/app/src/analytics/operations.ts
vincanger: ok I get it now. You want `statsQuery.orderBy.date` to be of type `desc` or `asc` and using `as const` is easier than typing the variable like `statsQuery: statsQueryInfo = {...}`

template/app/src/demo-ai-app/operations.ts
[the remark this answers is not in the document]
vincanger: yeah that was me being ignorant. keep assuming it's valid.

template/app/src/demo-ai-app/operations.ts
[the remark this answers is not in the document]
vincanger: makes sense. thanks!

template/app/src/demo-ai-app/operations.ts
vincanger: I'd say decrement after. They could spam a bit but wouldn't get very far. I like keeping it simple for the purposes of an example and giving the user the benefit of the doubt.

vincanger (approved):
Thanks for the detailed responses. Looks good.

template/app/src/demo-ai-app/operations.ts
sodic: Great, that's what we did. And the potential improvements are documented here: #421

PR #398, opened 2025-03-20
Adds Posthog script to opensaas.sh and docs.opensaas.sh

infomiho:


opensaas-sh/blog/astro.config.mjs
```diff
@@ -30,19 +30,10 @@ export default defineConfig({
         {
           tag: 'script',
           attrs: {
-            src: 'https://www.googletagmanager.com/gtag/js?id=G-8QGM76GR3Q',
+            defer: true,
+            src: '/piggy.js',
           },
         },
-        {
```
infomiho, line 36: I've removed the Google Analytics tracking code since we are not using them anymore IIRC

opensaas-sh/app_diff/src/server/scripts/dbSeeds.ts.diff
```diff
@@ -1,6 +1,6 @@
 --- template/app/src/server/scripts/dbSeeds.ts
 +++ opensaas-sh/app/src/server/scripts/dbSeeds.ts
-@@ -36,9 +36,11 @@
+@@ -37,9 +37,11 @@
```
infomiho, line 3: This was some minor diff update for the dbSeeds.ts file so I added it here as well

vincanger (approved):
looking gooood

PR #425, opened 2025-05-08
[Demo App] decrement credits only if user doesn't have subscription plan

### LINKED ISSUE #418: [Demo AI app] Only increment credits if there isn't an active subscription
When generating a GPT response, the demo AI app checks whether a user is subscribed with a flat-rate subscription or has remaining credits.

If the user is eligible, it subtracts their credits in both cases. It should instead only spend credits if there isn't an active flat-rate subscription.

vincanger:
## Description

Fixes #418

### DIFF
--- a/template/app/src/demo-ai-app/operations.ts
+++ b/template/app/src/demo-ai-app/operations.ts
@@ -74,7 +74,8 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput, 
     },
   });
 
-  const transactions: PrismaPromise<User | GptResponse>[] = [createResponse];
+  const transactions = [];
+  transactions.push(createResponse);
 
   if (!isUserSubscribed(context.user)) {
     const decrementCredit = context.entities.User.update({
--- a/template/app/src/demo-ai-app/operations.ts
+++ b/template/app/src/demo-ai-app/operations.ts
@@ -57,16 +57,24 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput, 
     throw new HttpError(500, 'Encountered a problem in communication with OpenAI');
   }
 
-  // We decrement the credits after using up tokens to get a daily plan
-  // from Chat GPT.
+  // We decrement the credits for users without an active subscription
+  // after using up tokens to get a daily plan from Chat GPT.
   //
   // This way, users don't feel cheated if something goes wrong.
   // On the flipside, users can theoretically abuse this and spend more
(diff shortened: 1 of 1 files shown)


template/app/src/demo-ai-app/operations.ts
```diff
@@ -81,18 +74,35 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
     },
   });
 
+  const transactions: PrismaPromise<User | GptResponse>[] = [createResponse];
```
sodic, line 77: Do you need the explicit type annotation here?

sodic (commented):
Nice, one small question.

template/app/src/demo-ai-app/operations.ts
sodic: Hmm, I guess it's subjective, but I prefer the way you did it anyway.

Although we might be better off moving everything up and doing it differently. See the other comment for more :)

template/app/src/demo-ai-app/operations.ts
```diff
@@ -57,22 +58,14 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
...
-  // Think about which option you prefer for you and edit the code accordingly.
-  const decrementCredit = context.entities.User.update({
-    where: { id: context.user.id },
-    data: {
-      credits: {
-        decrement: 1,
-      },
-    },
-  });
+  // Think about which option you prefer for your app and edit the code accordingly.
```
sodic: This comment is now in the wrong place I think

template/app/src/demo-ai-app/operations.ts
```diff
@@ -81,18 +74,35 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
...
+  if (!isUserSubscribed(context.user)) {
+    const decrementCredit = context.entities.User.update({
+      where: { id: context.user.id },
+      data: {
+        credits: {
+          decrement: 1,
+        },
+      },
+    });
+    transactions.push(decrementCredit);
```
sodic: How about we do this right away, instead of checking the credits and the subscription twice.

This is what I mean:
```ts
const transactions = [];
if (!isUserSubscribed(user)) {
  if (user.credits > 0) {
    const decrementCredit = ...
    transactions.push(decrementCredit);
  } else {
    throw new HttpError(402, 'User has not paid or is out of credits');
  }
}
```
We could then kick out the `isEligibleForResponse` function.

PR #426, opened 2025-05-12
Fix MacOS tools alias error 

### LINKED ISSUE #416: `./tools/dope./sh` fails for MacOS
I've followed readme's MacOS instructions for `opensaas-sh` but running the `patch.sh` fails when it calls the `dope.sh` script.

Error message:
```bash
patch: unrecognized option `--merge'
usage: patch [-bCcEeflNnRstuv] [-B backup-prefix] [-D symbol] [-d directory]
             [-g vcs-option] [-F max-fuzz] [-i patchfile] [-o out-file]
             [-p strip-count] [-r rej-name] [-T | -Z]
             [-V t | nil | never | none] [-x number] [-Y prefix]
             [-z backup-ext] [--quoting-style style] [--posix]
             [origfile [patchfile]]
       patch <patchfile

[Failed with exit code 1]
```

On my latests favicon changes PR I'm doing I tried to just remove the flag, it didn't break anything:
```bash
        patch_output=$(patch --no-backup-if-mismatch --merge "${DERIVED_DIR}/${derived_filepath}" < "${diff_filepath}")
```
->
```bash
        patch_output=$(patch --no-backup-if-mismatch "${DERIVED_DIR}/${derived_filepath}" < "${diff_filepath}")
```

Though, that might not be case for all situations.
We need to explore an alternative.

FranjoMindek:
Fixes problems caused by:
> Aliases and functions do not propagate to shell scripts

Old alias instructions are removed, instead we "alias" the command inside of the script itself.
The script exits if there is no `gpatch` on MacOS.
 
 Fixes #416

### DIFF
--- a/opensaas-sh/README.md
+++ b/opensaas-sh/README.md
@@ -26,18 +26,6 @@ If you're running the `patch.sh` or `diff.sh` scripts on Mac, you need to instal
 - `gpatch`,
 - and `diffutils`.
 
-You should also create aliases for `realpath` and `patch`:
-
-```sh
-brew install coreutils # contains grealpath
-brew install gpatch
-brew install diffutils
-
-echo 'alias realpath="grealpath"' >> ~/.zshrc
-echo 'alias patch="gpatch"' >> ~/.zshrc
-source ~/.zshrc
-```
-
 Make sure not to commit `app/` to git. It is currently (until we resolve this) not added to .gitignore because that messes up diffing for us.
 
 ### Blog (blog/)
--- a/opensaas-sh/README.md
+++ b/opensaas-sh/README.md
@@ -26,6 +26,12 @@ If you're running the `patch.sh` or `diff.sh` scripts on Mac, you need to instal
 - `gpatch`,
 - and `diffutils`.
 
+```sh
+brew install coreutils # contains grealpath
+brew install gpatch
+brew install diffutils
+```
+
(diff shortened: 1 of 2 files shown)


opensaas-sh/README.md
```diff
@@ -26,18 +26,6 @@ If you're running the `patch.sh` or `diff.sh` scripts on Mac, you need to instal
 - `gpatch`,
 - and `diffutils`.
 
-You should also create aliases for `realpath` and `patch`:
-
-```sh
-brew install coreutils # contains grealpath
-brew install gpatch
-brew install diffutils
```
infomiho, line 34: What about the `diff` utils? I believe they should also be installed with brew to get the correct behavior?

What about `grealpath`?

FranjoMindek: I should return install code block. It was part of aliases so I automatically discarded it.
Above is still mentioned to install those deps though.

FranjoMindek: It's back.

infomiho: Just to double check that `realpath` alias is no longer needed?

FranjoMindek: I tested this without any aliases and it worked. Aliases don't work in shell scripts anyway. That is the reason this PR started:

> Aliases and functions do not propagate to shell scripts

infomiho (approved):
LGTM

Issue #442, opened 2025-07-02
Update minimum node requirements to 20


FranjoMindek:
Following the wasps minimum node requirements.

PR #443, opened 2025-07-02
Update minimum node version to 20

FranjoMindek:
Fixes: https://github.com/wasp-lang/open-saas/issues/442

### DIFF
--- a/.github/workflows/blog-deployment.yml
+++ b/.github/workflows/blog-deployment.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: '20'
+          node-version: 'lts/*'
 
       - name: Install dependencies
         working-directory: ./opensaas-sh/blog
@@ -47,7 +47,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: '20'
+          node-version: 'lts/*'
 
       - name: Install dependencies
         working-directory: ./opensaas-sh/blog
--- a/.github/workflows/blog-deployment.yml
+++ b/.github/workflows/blog-deployment.yml
@@ -17,58 +17,58 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request'
     steps:
-    - name: Checkout code
-      uses: actions/checkout@v3
+      - name: Checkout code
+        uses: actions/checkout@v3
 
-    - name: Setup Node.js
-      uses: actions/setup-node@v3
-      with:
-        node-version: '18'
+      - name: Setup Node.js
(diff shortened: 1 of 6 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 env:
   WASP_TELEMETRY_DISABLE: 1
-  WASP_VERSION: 0.16.4
```
FranjoMindek, line 13: This one is extra, but we might as well update?

template/e2e-tests/package.json
```diff
@@ -29,6 +29,6 @@
     "prisma": "5.19.1"
   },
   "devDependencies": {
-    "@types/node": "^18.0.0"
```
FranjoMindek, line 32: Not related to template or wasp,
we could actually have this at v24 if we wanted to.

Edit: maybe best to leave it at 20 so it follows wasp's minimum req

.github/workflows/blog-deployment.yml
```diff
@@ -17,58 +17,58 @@ jobs:
...
     steps:
-    - name: Checkout code
-      uses: actions/checkout@v3
+      - name: Checkout code
+        uses: actions/checkout@v3
 
-    - name: Setup Node.js
-      uses: actions/setup-node@v3
-      with:
-        node-version: '18'
```
FranjoMindek, line 26: We could have this one at `lts/*` since its not related to template or wasp.

.github/workflows/blog-deployment.yml
```diff
@@ -17,58 +17,58 @@ jobs:
...
```
FranjoMindek, line 50: Same as above. `lts/*` possible.

.github/workflows/blog-deployment.yml
sodic: Is there a reason not to change it now?

FranjoMindek: Just wanted somebody else to agree with me on this.

FranjoMindek:
I've `grep`'d for both `node` and `18` separately.
Left over `package-lock.json` don't seem to be something we can change.
It's mostly:
```ts
"engines": {
  "node": ">=18"
}
```

I'm then gonna bump blog to `lts/*` ~~and e2e to `24`.~~
EDIT: not gonna bump e2e, because while it's not wasp project code, people might be using node@20 with wasp, so it seems weird to force v24 types on e2e, i'll leave it on v20 so it follows wasp

sodic:
> EDIT: not gonna bump e2e, because while it's not wasp project code, people might be using node@20 with wasp, so it seems weird to force v24 types on e2e, i'll leave it on v20 so it follows wasp

Good thinking!

sodic (approved):

PR #448, opened 2025-07-10
Updates for Wasp 0.17.0

infomiho:


opensaas-sh/app_diff/migrations/20250710084843_remove_newsletter_fields/migration.sql.diff
```diff
@@ -0,0 +1,11 @@
+--- template/app/migrations/20250710084843_remove_newsletter_fields/migration.sql
++++ opensaas-sh/app/migrations/20250710084843_remove_newsletter_fields/migration.sql
```
infomiho, line 2: There was a Prisma file change but no migration file was generated in the diffs so I added it

infomiho:
CI passes with the `0.17.0-rc4` https://github.com/wasp-lang/open-saas/actions/runs/16190716737/job/45705649989?pr=448

It will fail with `0.17.0` since it's not released yet, but that's okay.

PR #472, opened 2025-07-31
Refactor file upload and toast notifications

vincanger:
## Description

Replaces react-hot-toast with a custom toast system using @radix-ui/react-toast, updating all usages and adding new UI components for toast and dialog. Refactors file upload to use a two-step process: first generating an S3 upload URL, then adding the file to the database, and adds file deletion support with confirmation dialog and S3 cleanup. Updates Prisma schema, removes unused fields, and cleans up navigation and admin settings page.

https://github.com/user-attachments/assets/aa5f524a-e274-4c8c-8ced-c2ac625b992d

## Contributor Checklist


template/app/src/hooks/use-toast.ts
```diff
@@ -0,0 +1,194 @@
...
+// Inspired by react-hot-toast library
+import * as React from "react"
+
+import type {
+  ToastActionElement,
+  ToastProps,
+} from "../components/ui/toast"
+
+const TOAST_LIMIT = 1
+const TOAST_REMOVE_DELAY = 1000000
```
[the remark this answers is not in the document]
vincanger: 16.7 minutes?! wtf is this bot on about?

template/app/src/file-upload/operations.ts
```diff
@@ -80,3 +74,56 @@ export const getDownloadFileSignedURL: GetDownloadFileSignedURL<
...
+});
+
+type DeleteFileInput = z.infer<typeof deleteFileInputSchema>;
+
+export const deleteFile: DeleteFile<DeleteFileInput, File> = async (args, context) => {
+  if (!context.user) {
+    throw new HttpError(401);
+  }
+
+  await context.entities.File.findUniqueOrThrow({
```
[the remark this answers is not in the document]
vincanger: this is a decent suggestion. I will use this but change the order of deletion so it deletes the file from S3 only if the DB deletion is successful.

template/app/src/file-upload/FileUploadPage.tsx
```diff
@@ -98,86 +135,143 @@ export default function FileUploadPage() {
...
+                        <span>
+                          File <strong>{fileToDelete.name}</strong> deleted.
+                        </span>
+                      ),
+                    });
+                    allUserFiles.refetch();
+                  } catch (error: any) {
+                    toast({
+                      title: 'Error',
+                      description: error.message || 'Error deleting file.',
```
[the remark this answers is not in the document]
vincanger: I could probably just use a generic message like 'Error deleting file.' and not worry about the error.message, right?

template/app/src/demo-ai-app/DemoAppPage.tsx
```diff
@@ -130,10 +132,30 @@ function NewTaskForm({ handleCreateTask }: { handleCreateTask: typeof createTask
...
-      window.alert('Error: ' + (err.message || 'Something went wrong'));
+      if (err.statusCode === 402) {
+        const toastId = toast({
+          title: '⚠️ You are out of credits!',
+          style: {
+            minWidth: '400px',
+          },
+          action: (
+            <Link to={routes.PricingPageRoute.to}>
+              <Button variant='default' onClick={() => toastId.dismiss()}>
```
FranjoMindek: There is a `<ToastAction>` component that should be used instead.

Only problem is I don't think you've adapted it to the current theme.

I don't know what process did Fran do but we need to do the same.

template/app/src/admin/elements/settings/SettingsPage.tsx
```diff
@@ -15,15 +14,10 @@ const SettingsPage = ({ user }: { user: AuthUser }) => {
   useRedirectHomeUnlessUserIsAdmin({ user });
 
   const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
-    // TODO add toast provider / wrapper
+    // TODO implement
```
FranjoMindek, line 19: Do you want to implement this now?

template/app/src/demo-ai-app/DemoAppPage.tsx
```diff
@@ -130,10 +132,30 @@ function NewTaskForm({ handleCreateTask }: { handleCreateTask: typeof createTask
         hours: todaysHours,
       });
       if (response) {
-        setResponse(response as unknown as GeneratedSchedule);
+        setResponse(response);
       }
     } catch (err: any) {
-      window.alert('Error: ' + (err.message || 'Something went wrong'));
+      if (err.statusCode === 402) {
+        const toastId = toast({
```
FranjoMindek: When I tested with `<ToastAction>`, the toast disappeared on click.

We don't need this `toastId` constant then and custom dismiss logic.

template/app/src/file-upload/FileUploadPage.tsx
```diff
@@ -98,86 +135,143 @@ export default function FileUploadPage() {
...
+                            >
+                              <Download className='h-5 w-5' />
+                            </Button>
+                            <Button
+                              onClick={() => setFileToDelete(file)}
+                              variant='outline'
+                              size='sm'
+                              aria-label='Delete file'
+                            >
+                              <Trash className='h-5 w-5 text-destructive' />
```
FranjoMindek: This suffers from the same problem we had on Demo App's `TaskCard` component.
The red is not too visible on the dark mode.

To fix the core of the issue, we should change the dark mode `--destructive` CCS variable to be a bit brighter.
If we do so we can also revert to using `destructive` color instead of `red-500` on `TaskCard`.

This is something we should get in contact with Fran about.

template/app/src/file-upload/operations.ts
```diff
@@ -37,19 +40,10 @@ export const createFile: CreateFile<
     userId: context.user.id,
   });
 
-  await context.entities.File.create({
```
FranjoMindek, line 47: Why have we moved file creation to a separate action, while `deleteFile` action does both (remove s3 file and remove file from the database).

Now we have inconsistencies.

I don't see the benefit of doing them separately.

template/app/src/hooks/use-toast.ts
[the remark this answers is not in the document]
FranjoMindek: The bot has a nice hunch but he is wrong here.
He thinks that this variable is "toast duration in milliseconds", which would mean that by default toast messages disappear after 2.77 hours. A bit long. Usually it is 5 seconds.

However this variable is used to ensure the toast existing animation. So false positive.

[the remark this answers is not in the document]
FranjoMindek: And yes, we have a default of 5 seconds.

FranjoMindek (changes requested):
I haven't checked the `open-saas` part yet, but there are some fixes needed in the template.

template/app/src/demo-ai-app/DemoAppPage.tsx
vincanger: I updated `ToastAction` to fit our styling so they look like this:
image
image

template/app/src/file-upload/FileUploadPage.tsx
vincanger: Good point. Made a sub-issue for it here https://github.com/wasp-lang/open-saas/issues/481

template/app/src/file-upload/operations.ts
vincanger: The handleUpload function in FileUploadPage.tsx component acts as the orchestrator for a single user action so that we don't accidentally add files to the DB that don't get uploaded to the S3 bucket. 
1. It asks the server for a signed URL (createFileUploadUrl).
2. It uploads the file to S3 (this is the part the user sees progress for).
3. It tells the server the work is done, and the server verifies and saves the record (addFileToDb).
    
What was happening before when everything was in one operation was that, in some cases e.g. if we were using the wrong credentials, we'd get a faulty URL returned and would have already saved the file key in the db. So the upload to the URL would fail, but we'd still get a record for that file in the db for it. 

I guess we could go ahead right away and add the record to the db with a flag and update the file after upload is complete, but then we'd still have the split logic so I'm not sure it's any better.

template/app/src/admin/elements/settings/SettingsPage.tsx
vincanger: We don't have any functionality for the settings page at the moment, so I think I'll just leave it for now, but I created an issue where we can consider how to implement it and/or the General User's account info as well >> https://github.com/wasp-lang/open-saas/issues/482

vincanger:
@FranjoMindek made a couple changes based on your comments and replied to others :)

template/app/src/demo-ai-app/operations.ts
```diff
@@ -39,6 +39,22 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
     throw new HttpError(401, 'Only authenticated users are allowed to perform this operation');
   }
 
+    if (!isUserSubscribed(context.user)) {
+      if (context.user.credits > 0) {
+        const decrementCredit = context.entities.User.update({
```
FranjoMindek: We don't use `decrementCredit` anywhere, and we don't await the update.

```ts
await context.entities.User.update({
  where: { id: context.user.id },
  data: {
    credits: {
      decrement: 1,
    },
  },
});
```

template/app/src/demo-ai-app/operations.ts
```diff
@@ -39,6 +39,22 @@ export const generateGptResponse: GenerateGptResponse<GenerateGptResponseInput,
...
+          where: { id: context.user.id },
+          data: {
+            credits: {
+              decrement: 1,
+            },
+          },
+        });
+
+      } else {
+        throw new HttpError(402, 'User has not paid or is out of credits');
```
FranjoMindek: Has not paid for what?
If I read it right, the user doesn't have an active subscription or is out of credits.

template/app/src/demo-ai-app/operations.ts
FranjoMindek, line 1: Now I see that we already have this piece of code below? Why the duplication?
Did you mean to move it?

template/app/src/file-upload/operations.ts
FranjoMindek: Why not just change the order of execution inside of that operation then?
We could still do it in a single operation.

We create and URL, we upload, on success we create a DB record.

template/app/src/demo-ai-app/DemoAppPage.tsx
FranjoMindek: On hover seems weird for dark mode?
image
image

template/app/src/file-upload/FileUploadPage.tsx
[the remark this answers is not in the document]
FranjoMindek: We should still use the message if it exists. 
But have a generic backup yes.

FranjoMindek (changes requested):
Some more changes and questions.

template/app/src/demo-ai-app/DemoAppPage.tsx
vincanger: You're right. Changed it to match the buttons on the hero section of the LP.

template/app/src/demo-ai-app/operations.ts
vincanger: whoops this was a leftover. I move that check up to the top so that it failed fast and I could test the toast behaviour without triggering any openai calls.

template/app/src/demo-ai-app/operations.ts
vincanger: see above

template/app/src/file-upload/operations.ts
```diff
@@ -37,22 +45,41 @@ export const createFile: CreateFile<
...
-    data: {
-      name: fileName,
-      key,
-      uploadUrl: s3UploadUrl,
-      type: fileType,
-      user: { connect: { id: context.user.id } },
-    },
-  });
-
   return {
```
FranjoMindek: We can just:

```ts
return await getUploadFileSignedURLFromS3({
  fileType,
  fileName,
  userId: context.user.id,
});
  ```

File: template/app/src/file-upload/operations.ts (hunk 8 above)
Comment (vincanger) [reply template/app/src/file-upload/operations.ts:47]: Here's why the first approach is preferred:
**Performance & User Experience**: Large file uploads can take time. In the single-operation model, the user's connection to the Wasp server must stay open for the entire duration of the upload from the client to the server, AND from the server to S3. With direct uploads, the client has a fast, dedicated connection to S3, a service optimized for this, and our server is freed up to handle other requests.
**Scalability & Cost**: By offloading the bandwidth-heavy task to S3, our Wasp server uses fewer resources (CPU, memory, network I/O) per upload. This means a single server instance can handle more concurrent users, and we're less likely to need to scale up our server infrastructure. In a serverless environment, this also translates directly to lower costs, since we're not paying for our server function to be running for the entire duration of the upload.
**Security**: Presigned URLs are temporary and scoped to a specific file and action. This is a very secure way to grant a client write-access to a specific location in our bucket without exposing our secret keys.
In conclusion, while the single-operation approach might seem simpler at first, it introduces significant performance, scalability, and cost issues, especially as the application grows. The two-operation, direct-to-S3 approach is a modern, robust pattern that leverages the strengths of each part of the system: the client for handling the transfer, S3 for storing the data, and our server for orchestration and business logic.

File: template/app/src/file-upload/FileUploadPage.tsx [hunk 12]
```diff
@@ -82,7 +100,26 @@ export default function FileUploadPage() {
         return;
       }
 
-      await uploadFileWithProgress({ file: file as FileWithValidType, setUploadProgressPercent });
+      const fileWithValidType = file as FileWithValidType;
```
FranjoMindek: We generally try to avoid asserts as this.
We did use `validateFile`, but its usually bad to disconnect verification from return type.
This is a bit of old code but:

I would first change `fileUploading,.ts`
```ts
// simplified a bit
export type FileWithValidType = File & { type: AllowedFileType };
// ...

export function validateFile(file: File): FileWithValidType {
  if (file.size > MAX_FILE_SIZE_BYTES) {
    throw new Error(`File size exceeds ${MAX_FILE_SIZE_BYTES / 1024 / 1024}MB limit.`);
  }

  if (!isFileWithAllowedFileType(file)) {
    throw new Error(`File type '${file.type}' is not supported.`);
  }

  return file;
}

function isFileWithAllowedFileType(file: File): file is FileWithValidType {
  return (ALLOWED_FILE_TYPES as readonly string[]).includes(file.type);
}
```

This way once we validate the file, we already have the correct return type.

Then I would propagate it here:
```ts
// we have string now
// we never used the code anyways
const [uploadError, setUploadError] = useState<string | null>(null);
// ...

const handleUpload = async (e: FormEvent<HTMLFormElement>) => {
  try {
    e.preventDefault();

    const formElement = e.target;
    if (!(formElement instanceof HTMLFormElement)) {
      throw new Error('Event target is not a form element');
    }

    const formData = new FormData(formElement);
    const formDataFileUpload = formData.get('file-upload');

    if (!formDataFileUpload || !(formDataFileUpload instanceof File)) {
      setUploadError('Please select a file to upload');
      return;
    }

    const file = validateFile(formDataFileUpload);

    const { s3UploadUrl, s3UploadFields, key } = await createFileUploadUrl({
      fileType: file.type,
      fileName: file.name,
    });

    await uploadFileWithProgress({
      file,
      s3UploadUrl,
      s3UploadFields,
      setUploadProgressPercent,
    });

    await addFileToDb({
      key,
      fileType: file.type,
      fileName: file.name,
    });

    formElement.reset();
    allUserFiles.refetch();
  } catch (error) {
    console.error('Error uploading file:', error);
    setUploadError(String(error));
  } finally {
    setUploadProgressPercent(0);
  }
};
```

template/app/src/file-upload/operations.ts
FranjoMindek: We need separate actions because we don't want to upload to server, as pointed out by Vinny.
I forgot this aspect.

template/app/src/file-upload/operations.ts
```diff
@@ -18,11 +25,12 @@ const createFileInputSchema = z.object({
 
 type CreateFileInput = z.infer<typeof createFileInputSchema>;
 
-export const createFile: CreateFile<
+export const createFileUploadUrl: CreateFileUploadUrl<
   CreateFileInput,
   {
     s3UploadUrl: string;
     s3UploadFields: Record<string, string>;
+    key: string;
```
FranjoMindek: Can we just rename this `key` to `s3Key` everywhere?

FranjoMindek (changes requested):
The new upload logic seems okay.

A bit comments on some older stuff.
Could go nice with current PR.

template/app/src/file-upload/FileUploadPage.tsx
vincanger: Good idea. I implemented this along with removing the `setUploadError` logic in favor of using the toasts, so I got rid of that `FileUploadError` type.

vincanger:
@FranjoMindek pushed some more requested fixes along with a couple cleanup items.

opensaas-sh/app_diff/package.json.diff
```diff
@@ -13,9 +13,9 @@
      "@aws-sdk/client-s3": "^3.523.0",
      "@aws-sdk/s3-presigned-post": "^3.750.0",
 @@ -36,6 +41,7 @@
+     "react-apexcharts": "1.4.1",
```
FranjoMindek, line 16: Where did `"react-apexcharts"` come from?

vincanger: We're still using `ReactApexChart`. in `template/app/src/admin/dashboards/analytics/RevenueAndProfitChart.tsx`

Here's a new issue for its refactor/removal: https://github.com/wasp-lang/open-saas/issues/485

infomiho:
One meta comment: it would be better if this was split into two PRs: one for toast and the other for upload logic. 

I'll review the logic now - but I see Franjo did a few iterations himself, nice.

opensaas-sh/app_diff/src/file-upload/workers.ts.diff
```diff
@@ -0,0 +1,32 @@
...
++++ opensaas-sh/app/src/file-upload/workers.ts
+@@ -0,0 +1,29 @@
++import type { DeleteFilesJob } from 'wasp/server/jobs';
++import { deleteFileFromS3 } from './s3Utils';
++
++export const deleteFilesJob: DeleteFilesJob<never, void> = async (_args, context) => {
++  const filesToDelete = await context.entities.File.findMany({
++    where: {
++      createdAt: {
++        lt: new Date(Date.now() - 1000 * 60 * 60 * 24 * 7),
```
infomiho: I'd suggest we do 
```ts
const last7Days = new Date(Date.now() - 1000 * 60 * 60 * 24 * 7);
```
So it's more clear what is the limit. Or even better:
```ts
const dayInMiliseconds = 1000 * 60 * 60 * 24;
const timeLimitToDeleteFiles = Date.now() - 7 * dayInMiliseconds;
```
Something along those lines.

opensaas-sh/app_diff/src/file-upload/workers.ts.diff
```diff
@@ -0,0 +1,32 @@
...
++    } catch (err) {
++      console.error(`Failed to delete S3 file with key ${file.s3Key}:`, err);
++    }
++  }
++
++  const deletedFiles = await context.entities.File.deleteMany({
++    where: {
++      id: { in: filesToDelete.map((file) => file.id) },
++    },
++  });
```
infomiho: This is fine, but it would be better done in a Prisma transaction:
```ts
try {
  await prisma.$transaction(async (tx) => {
    await tx.file.delete(...)
    await deleteFileFromS3(...)
  })
} catch (err) {
  // If something went wrong with deleting the file from S3, the file won't be deleted from the database
}
```

I'm not sure if we this is too fancy @FranjoMindek

template/app/src/file-upload/FileUploadPage.tsx
```diff
@@ -66,118 +82,189 @@ export default function FileUploadPage() {
...
+                    const errorMessage = error instanceof Error ? error.message : 'Error deleting file.';
+                    toast({
+                      title: 'Error',
+                      description: errorMessage,
+                      variant: 'destructive',
+                    });
+                  } finally {
+                    setFileToDelete(null);
+                  }
+                }}
```
infomiho: Let's move this into a proper function and not inline it, it's a lot of logic to be inline IMHO

template/app/src/file-upload/operations.ts
```diff
@@ -69,14 +90,43 @@ export const getAllFilesByUser: GetAllFilesByUser<void, File[]> = async (_args,
...
+        id: context.user.id,
+      },
+    },
+  });
+
+  try {
+    await deleteFileFromS3({ s3Key: deletedFile.s3Key });
+  } catch (error) {
+    console.error(`S3 deletion failed. Orphaned file s3Key: ${deletedFile.s3Key}`, error);
+  }
```
infomiho, line 151: Prisma transaction would be also useful here. If the s3 deletion fails, don't delete it from the database. @FranjoMindek is this an overkill to do here?

template/app/src/file-upload/s3Utils.ts
```diff
@@ -20,29 +20,53 @@ type S3Upload = {
...
+
+export const checkFileExistsInS3 = async ({ s3Key }: { s3Key: string }) => {
+  const command = new HeadObjectCommand({
+    Bucket: process.env.AWS_S3_FILES_BUCKET,
+    Key: s3Key,
+  });
+  try {
+    await s3Client.send(command);
+    return true;
+  } catch (error: any) {
```
infomiho: Can we do better than using `any` here? Let's use `instanceof` to check if `unknown` is some class from S3.

template/app/src/hooks/use-toast.ts
```diff
@@ -0,0 +1,194 @@
+"use client"
```
infomiho: Wasp doesn't need these for client components. It might prompt our users to think it's necessary. Let's remove it.

opensaas-sh/app_diff/src/file-upload/workers.ts.diff
vincanger: We use prisma transactions elsewhere in the code so I think its fine to use it here too.

template/app/src/hooks/use-toast.ts
vincanger: Whoops. Good catch.

opensaas-sh/app_diff/src/file-upload/workers.ts.diff
vincanger: I was thinking this approach instead, and Claude said this would be better than prisma transactions because of performance, and that it would pick up orphaned file deletions on next run:

```ts
const filesToDelete = await context.entities.File.findMany({
    where: {
      createdAt: {
        lt: new Date(sevenDaysAgo),
      },
    },
    select: { s3Key: true, id: true },
  });

  const deletionResults = await Promise.allSettled(
    filesToDelete.map(async (file) => {
      await deleteFileFromS3({ s3Key: file.s3Key });
      return file.id;
    })
  );

  const successfullyDeletedFromS3Ids = deletionResults
    .filter((result) => result.status === 'fulfilled')
    .map((result) => result.value);

  const deletedFiles = await context.entities.File.deleteMany({
    where: {
      id: { in: successfullyDeletedFromS3Ids },
    },
  });
  ```

File: template/app/src/file-upload/operations.ts (hunk 18 above)
Comment (vincanger) [reply template/app/src/file-upload/operations.ts:151]: Some research says that it's bad practice to pack external API calls / services into a prisma transaction because it can lock any other db execution if the service takes too long to respond or has a problem.

Also, if the service call, in this case S3 deletion works, but the DB call fails, then there is no way of turning back the S3 deletion leaving you with the same issue.

So I went for a different approach using a cleanup job. We still delete the file from our db, but that means file deletions from s3 could fail and leave orphaned files. So I added a cleanup job that finds keys that exist in the S3 bucket but don't exist in our db and deletes those orphaned files from s3.

File: template/app/src/file-upload/workers.ts [hunk 21]
```diff
@@ -0,0 +1,65 @@
+import type { CleanUpOrphanedFilesS3Job } from 'wasp/server/jobs';
```
infomiho: This is pretty advanced stuff but it might be problematic if the user has e.g. 10000 files in S3 and we fetch 10000 IDs and then we fetch all the files from the DB and compare. It's probably inefficient, that's what I'm trying to say.

If we don't do it, there might be some orphaned files on S3 - I guess that's not that dramatic, but I like that we are trying to solve it.

I'm on the fence if we should do this or just go for the simplest kind of solution of doing nothing.

infomiho: What I'm saying is that we can't offer the user a 100% bullet proof, optimized solution out of the box. But it's better to give them something they can optimize if they need to?

vincanger:
@infomiho and I talked about the cleanup job and decided it would be best to remove it and add a note about it in the docs in case users want to implement it themselves.

infomiho:
@vincanger let's fix the conflict and merge 👍

infomiho (approved):

PR #477, opened 2025-08-01
Upgrade to Node 22.12

cprecioso:
Following 
- https://github.com/wasp-lang/wasp/issues/2913

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -24,7 +24,7 @@ jobs:
         id: setup-node
         uses: actions/setup-node@v4
         with:
-          node-version: '20'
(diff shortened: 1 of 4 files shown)


vincanger (approved):
noice. approved!

cprecioso:
Yeah, I thought we had a `main`/`release` situation here too

FranjoMindek:
We do have tags, which are used by the Wasp CLI for the starter template.
We should be able to merge this now.

https://github.com/wasp-lang/wasp/blob/1c56fdef227a9c3f2036fce3ac655dd5c3b5b22b/waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs#L189-L194

PR #491, opened 2025-08-08
Optimize the NavBar Scaling using `will-change-transform`

fgrgic:
## Description

Replaced throttled scroll handlers with GPU-accelerated CSS transforms for smoother animations.
Changes: 
- Removed: `throttleWithTrailingInvocation` utility (no longer used)
- Replaced: Throttled scroll handler with direct event listener
- Updated: All animations to use transform scale instead of size/font changes
- Added: will-change-transform for GPU acceleration

Why
`will-change-transform` tells the browser to create a separate GPU layer for an element, so CSS transforms run smoothly without expensive layout recalculations.

## Contributor Checklist


template/app/src/client/components/NavBar/NavBar.tsx
```diff
@@ -23,22 +22,26 @@ export default function NavBar({ navigationItems }: { navigationItems: Navigatio
...
     };
   }, []);
 
   return (
     <>
       {isLandingPage && <Announcement />}
-      <header className={cn('sticky top-0 z-50 transition-all duration-300', isScrolled && 'top-4')}>
+      <header
+        className={cn(
+          'sticky top-0 z-50 transition-transform duration-300 ease-out will-change-transform',
```
FranjoMindek, line 41: `will-change-transform` should be used for performance reasons.
Our animations are really simple in nature.

I don't think our scroll problems are due to performance problems.
I don't think we need this.

You are free to discuss this.

FranjoMindek: More info: https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

PR #493, opened 2025-08-11
Strapi and payment processing improvements

FranjoMindek:
### Changes:

`fetchStripeCustomer`
- Bad name / bad behavior
	- The function can also create a new customer which isn't implied by name
- Bad logic? (didn't change anything here for now)
	- `email` is not a unique field for `Stripe.Customer`
	- Currently we only allow 1 `Stripe.Customer` per 1 `User.email` but that is enforced by our own application logic
	- We already store `Stripe.Customer.id` under `User.paymentProcessorUserId`
	- We could easily search `Stripe.Customer` by its `id`, which is unique, rather by its `email`   

`stripePaymentProcessor.createCheckoutSession`
- Bad logic
	- We:
		- fetch/create a `Stripe.Customer`
		- create a `Stripe.Checkout.Session`
		- connect `User` to `Stripe.Customer`  
	- If we are creating a `Stripe.Customer` (rather than fetching), and we error on creating a `Stripe.Checkout.Session`, we won’t connect `User` to the newly created `Stripe.Customer`
	- However due to our application logic, if we retry, we will re-fetch the same `Stripe.Customer` since we fetch it by `email`

`paymentDetails.ts` -> `userPaymentDetails.ts`
- Bad name
	- This module is solely responsible for updating `User` entity, let's mention the user in the name

`stripePaymentProcessor.fetchCustomerPortalUrl`
- Bad usage 
	- We use a billing portal URL which isn't a secret. We set is as `STRIPE_CUSTOMER_PORTAL_URL` and then we fetch it though server on the client. 
		- The users still have to login to their Stripe customer account to use it even though they are logged in the application.
		- We can easily use the link directly on the client
	- I think we meant to use `stripeClient.billingPortal.sessions.create`, which creates a billing portal URL for that specific customer
		- The user is already logged in to their Stripe customer account inside of that link
		- I've implemented this in the PR, but **we might not want this behavior**, do let me know if that is true
			- Feel free to test it out in the template
			- TODO: if we go this route 
				- update the docs to reflect the new billing portal
				- remove `STRIPE_CUSTOMER_PORTAL_URL`

We parse the Stripe payload with our custom `zod` schemas.
We don't have to do that since we already confirm the data via `stripeClient.webhooks.constructEvent`.
Plus we lose proper type information through our `zod` schemas.
I deleted the Stripe `webhookPayload.ts`.

Payment details auditing problems
- We always overwrite user payment details, instead we should save each one of them as a separate record
- Didn't fix this

Stripe now supports creating invoices for one-time payments, so I refactored our Stripe's `webhook.ts` around that.
That means we handle all of the initial payment effects inside of the `invoice.paid` event.
- **Someone with access to Stripe will need to update our webhook settings, we no longer need the `checkout.session.completed` events**

Stripe's old plan id is now called price id, we should update env vars to reflect that, as not to confuse the template users:
- `PAYMENTS_HOBBY_SUBSCRIPTION_PLAN_ID` -> `PAYMENTS_HOBBY_SUBSCRIPTION_PRICE_ID`
- `PAYMENTS_PRO_SUBSCRIPTION_PLAN_ID` -> `PAYMENTS_PRO_SUBSCRIPTION_PRICE_ID`
- `PAYMENTS_CREDITS_10_PLAN_ID` -> `PAYMENTS_CREDITS_10_PRICE_ID`

`CheckoutPage.tsx` -> `CheckoutResultPage.tsx`
- We use it to say success or failure, not actual checkout stuff

Manage subscription button:
- Needs to reflect other changes
	- Since one-time payments now generate invoices, we want all users which have a `Stripe.Customer` account to be able to access their billing portal, before only users with subscription could access it

Other than mentioned changes I've also clean up the types and general logic to be clearer.


template/app/src/payment/stripe/userPaymentDetails.ts
```diff
@@ -0,0 +1,59 @@
+import { PrismaClient } from '@prisma/client';
+import Stripe from 'stripe';
+import type { SubscriptionStatus } from '../plans';
+import { PaymentPlanId } from '../plans';
+
+export async function updateUserStripePaymentDetails(
```
FranjoMindek, line 6: This is just old `updateUserStripePaymentDetails` from `stripe/paymentDetails`, but I've split it up into two separate functions which correspond to two different actions it does.

template/app/src/payment/stripe/stripeClient.ts
```diff
@@ -1,7 +1,7 @@
 import Stripe from 'stripe';
 import { requireNodeEnvVar } from '../../server/utils';
 
-export const stripe = new Stripe(requireNodeEnvVar('STRIPE_API_KEY'), {
```
FranjoMindek, line 4: Having it named `stripe` makes it easy to confuse with the `Stripe` namespace often imported everywhere.
And it is a client.

opensaas-sh/app_diff/.env.vault.diff
vincanger, line 1: What env vars were changed here? I want to make sure we're still using the old Product IDs and vars from Stripe.

opensaas-sh/app_diff/fly-client.toml.diff
vincanger, line 1: any idea why this changed? I assume you didn't actually deploy anything.

template/app/src/shared/utils.ts
vincanger, line 1: nice

vincanger (changes requested):
Cool. Looks great. Just a couple comments + we need to update docs.

opensaas-sh/app_diff/fly-client.toml.diff
FranjoMindek: The diff got applied twice to the same file, it's just duplicated. I'll revert.

opensaas-sh/app_diff/.env.vault.diff
FranjoMindek: Same as for `fly-client.toml` I've applied the diff twice with scripts, my bad.

FranjoMindek:
Updated docs. Looks like we didn't generate llms text for previous versions so there is some extra.

FranjoMindek:
Closed in favor of:

https://github.com/wasp-lang/open-saas/pull/505
https://github.com/wasp-lang/open-saas/pull/506
https://github.com/wasp-lang/open-saas/pull/507

Issue #483, opened 2025-08-05
use Wasp's new Custom Env Var validation schema for validating integrations


vincanger:
We currently use a custom env var validation setup but should refactor everything and add validation using Wasp's custom env var vailidation logic, e.g. :

```ts
  server: {
    envValidationSchema: import { envValidationSchema } from "@src/server/validation",
  },
```

```ts
// `src/file-upload/validation.ts`

export const fileUploadEnvValidationSchema = z.object({
  AWS_S3_REGION: z.string({ required_error: 'AWS_S3_REGION is required' }),
  AWS_S3_FILES_BUCKET: z.string({ required_error: 'AWS_S3_FILES_BUCKET is required' }),
  AWS_S3_IAM_ACCESS_KEY: z.string({ required_error: 'AWS_S3_IAM_ACCESS_KEY is required' }).regex(/^[A-Z0-9]+$/, 'Invalid AWS IAM access key'),
  AWS_S3_IAM_SECRET_KEY: z.string({ required_error: 'AWS_S3_IAM_SECRET_KEY is required' }).regex(/^[A-Za-z0-9/+=]+$/, 'Invalid AWS IAM secret key'),
});
```

```ts
// `src/server/validation.ts

import { defineEnvValidationSchema } from 'wasp/env';
import { fileUploadEnvValidationSchema } from '../file-upload/validation';

export const envValidationSchema = defineEnvValidationSchema(z.object({
  ...fileUploadEnvValidationSchema.shape,
}));
```

```bash
AWS_S3_IAM_ACCESS_KEY=EXAMPLEACCE55KEY
AWS_S3_IAM_SECRET_KEY=example+Secr3tK3y
AWS_S3_FILES_BUCKET=your-bucket-name
AWS_S3_REGION=your-region
```

fixes #370

PR #505, opened 2025-09-17
Improve Stripe payment provider logic

FranjoMindek:
## Description

> Describe your PR! If it fixes specific issue, mention it with "Fixes # (issue)".

## Contributor Checklist


opensaas-sh/blog/public/llms-full.txt
```diff
@@ -23,7 +23,8 @@ If you find this template useful, consider giving us [a star on GitHub](https://
 ## What's inside?
 
 The template itself is built on top of some very powerful tools and frameworks, including:
```
FranjoMindek: This hasn't been run in a while so more changes pop up.

template/app/src/payment/plans.ts
```diff
@@ -35,7 +38,7 @@ export const paymentPlans: Record<PaymentPlanId, PaymentPlan> = {
     getPaymentProcessorPlanId: () => requireNodeEnvVar('PAYMENTS_CREDITS_10_PLAN_ID'),
     effect: { kind: 'credits', amount: 10 },
   },
-};
+} as const satisfies Record<PaymentPlanId, PaymentPlan>;
```
FranjoMindek, line 46: Allows us to do e.g. `paymentPlans.credits10.effect.amount` which would throw before.

template/app/src/shared/utils.ts
```diff
@@ -2,8 +2,7 @@
  * Used purely to help compiler check for exhaustiveness in switch statements,
  * will never execute. See https://stackoverflow.com/a/39419171.
  */
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-export function assertUnreachable(x: never): never {
+export function assertUnreachable(_: never): never {
```
FranjoMindek, line 5: Just a quick fix.

template/app/src/payment/stripe/stripeClient.ts
```diff
@@ -1,12 +1,30 @@
...
+ *
+ * You can override this by setting `apiVersion` when creating the Stripe client.
+ * This is useful for testing new API versions before updating the
+ * default version in your Dashboard.
+ *
+ * The Stripe Node SDK works with multiple API versions and follows semantic versioning.
+ * Major SDK versions typically correspond to Stripe's biannual releases (like 'basil').
+ * Each SDK version uses the API version that was current when it was released.
+ *
+ * Note: '2025-04-30.basil' follows Stripe's newer versioning format where:
```
FranjoMindek, line 17: As opposed to older date only formats.
Maybe word differently?

template/app/src/payment/stripe/webhook.ts
```diff
@@ -1,56 +1,59 @@
...
+        }
     }
-    return response.json({ received: true }); // Stripe expects a 200 response to acknowledge receipt of the webhook
+    return response.status(204).send();
   } catch (err) {
     if (err instanceof UnhandledWebhookEventError) {
-      console.error(err.message);
-      return response.status(422).json({ error: err.message });
+      response.status(204).send();
+      throw err;
```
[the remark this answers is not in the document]
FranjoMindek: Sending `422` here would make `stripe` re-send the same "unhandled" webhook event again on exponential back-off basis.
If the serve can automatically recover, this would just crash the server again.


I think we should crash and acknowledge the event so `strip` stops sending it, as this is unexpected behavior and needs further manual webhook configuration.

template/app/src/payment/stripe/checkoutUtils.ts
```diff
@@ -1,64 +1,69 @@
...
-const DOMAIN = process.env.WASP_WEB_CLIENT_URL || 'http://localhost:3000';
+const CLIENT_BASE_URL = process.env.WASP_WEB_CLIENT_URL || 'http://localhost:3000';
 
-export async function fetchStripeCustomer(customerEmail: string) {
-  let customer: Stripe.Customer;
+/**
+ * Returns a Stripe customer for the given User email, creating a customer if none exist.
+ * Implements email uniqueness logic since Stripe doesn't enforce unique emails.
+ */
+export async function ensureStripeCustomer(userEmail: string): Promise<Stripe.Customer> {
```
FranjoMindek: Old name `fetchStripeCustomer` lied, because it also created a customer if none existed.

template/app/src/payment/stripe/checkoutUtils.ts
```diff
@@ -1,64 +1,69 @@
...
     console.error(error);
     throw error;
   }
 }
 
 interface CreateStripeCheckoutSessionParams {
-  priceId: string;
-  customerId: string;
-  mode: StripeMode;
+  priceId: Stripe.Price['id'];
```
FranjoMindek: I try to extract types from `stripe` as much as I can.
If I miss it anywhere it is a mistake.

template/app/src/payment/stripe/checkoutUtils.ts
```diff
@@ -1,64 +1,67 @@
...
-      customer: customerId,
-    });
-  } catch (error) {
-    console.error(error);
-    throw error;
-  }
+    ],
+    mode,
+    success_url: `${CLIENT_BASE_URL}/checkout?status=success`,
+    cancel_url: `${CLIENT_BASE_URL}/checkout?status=cancel`,
```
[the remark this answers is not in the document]
FranjoMindek: Good one.

template/e2e-tests/package.json
```diff
@@ -12,14 +12,14 @@
     "_comment-on-local:e2e:cleanup-stripe": "NOTE: because we are running the stripe webhook listener in the background, we want to make sure we kill the previous processes before starting a new one.",
     "e2e:playwright": "DEBUG=pw:webserver npx playwright test",
     "local:e2e:cleanup-stripe": "PID=$(ps -ef | grep 'stripe listen' | grep -v grep | awk '{print $2}') || true && kill -9 $PID || true",
-    "local:e2e:playwright:ui": "npx playwright test --ui",
+    "local:e2e:playwright:ui": "SKIP_EMAIL_VERIFICATION_IN_DEV=true npx playwright test --ui",
```
FranjoMindek, line 15: Explore: how come we don't set this by default inside of `wasp-app-runner` if we depend on it?

Why did I add this? Did I do something wrong?

template/app/src/payment/stripe/checkoutUtils.ts
```diff
@@ -1,64 +1,67 @@
...
-      ],
-      mode: mode,
-      success_url: `${DOMAIN}/checkout?status=success`,
-      cancel_url: `${DOMAIN}/checkout?status=canceled`,
-      automatic_tax: { enabled: true },
-      allow_promotion_codes: true,
-      customer_update: {
-        address: "auto",
+}: CreateStripeCheckoutSessionParams): Promise<Stripe.Checkout.Session> {
+  // WASP_WEB_CLIENT_URL will be set up by Wasp when deploying to production: https://wasp.sh/docs/deploying
```
FranjoMindek: Reminder to create a issue about solving the env vars situation on `open-saas`.
Use Wass's `defineEnvValidationSchema`.
Centralize them in a single place.

vincanger: this one already exists here https://github.com/wasp-lang/open-saas/issues/483

template/e2e-tests/package.json
infomiho: The idea of `wasp-app-runner` is to run the app and not concern itself with env variables. This was suggested by @sodic and accepted by me so it's following more the Unix philosophy of simple tools doing a single job. That's why the `wasp-app-runner` "only" runs Wasp apps.

opensaas-sh/blog/src/content/docs/guides/deploying.mdx
```diff
@@ -175,7 +175,7 @@ export const stripe = new Stripe(process.env.STRIPE_KEY!, {
 2. click on `+ add endpoint`
 3. enter your endpoint url, which will be the url of your deployed server + `/payments-webhook`, e.g. `https://open-saas-wasp-sh-server.fly.dev/payments-webhook`
 <Image src={stripeListenEvents} alt="listen events" loading="lazy" />
-4. select the events you want to listen to. These should be the same events you're consuming in your webhook which you can find listed in [`src/payment/stripe/webhookPayload.ts`](https://github.com/wasp-lang/open-saas/blob/main/template/app/src/payment/stripe/webhookPayload.ts):
+4. select the events you want to listen to. These should be the same events you're consuming in your webhook which you can find handled in [`src/payment/stripe/webhook.ts`](https://github.com/wasp-lang/open-saas/blob/main/template/app/src/payment/stripe/webhook.ts):
```
infomiho:
```suggestion
4. select the events you want to listen to. These should be the same events you're consuming in your webhook which are handled in [`src/payment/stripe/webhook.ts`](https://github.com/wasp-lang/open-saas/blob/main/template/app/src/payment/stripe/webhook.ts):
```

opensaas-sh/blog/src/content/docs/guides/payments-integration.mdx
```diff
@@ -95,22 +95,21 @@ To create a test product, go to the test products url [https://dashboard.stripe.
 
 ### Create a Test Customer
 
-To create a test customer, go to the test customers url [https://dashboard.stripe.com/test/customers](https://dashboard.stripe.com/test/customers).
+You can create a test customer directly in the [Stripe Dashboard](https://dashboard.stripe.com/test/customers).
```
infomiho:
```suggestion
You can create a test customer in the [Stripe Dashboard](https://dashboard.stripe.com/test/customers).
```

Nitpick: I feel like it's cleaner

opensaas-sh/blog/src/content/docs/guides/payments-integration.mdx
```diff
@@ -95,22 +95,21 @@ To create a test product, go to the test products url [https://dashboard.stripe.
...
 
 - Click on the `Add a customer` button and fill in the relevant information for your test customer.
-:::note
- When filling in the test customer email address, use an address you have access to and will use when logging into your SaaS app. This is important because the email address is used to identify the customer when creating a subscription and allows you to manage your test user's payments/subscriptions via the test customer portal
-:::
+
+Alternatively, Open SasS will automatically create a test customer the first time a user starts a checkout session. 
+This customer is linked to the email address associated with your app's user.
```
infomiho:
```suggestion
Alternatively, Open SasS automatically creates a test customer the first time a user starts a checkout session. 
This customer is linked to the email address associated with the user in your app.
```
I think it reads better if the tense is consistent i.e. we keep using present.

opensaas-sh/blog/src/content/docs/guides/payments-integration.mdx
```diff
@@ -95,22 +95,21 @@ To create a test product, go to the test products url [https://dashboard.stripe.
...
 ### Set up the Customer Portal
 
-Go to https://dashboard.stripe.com/test/settings/billing/portal in the Stripe Dashboard and activate and copy the `Customer portal link`. Paste it in your `.env.server` file:
+You can set up your customer portal in your [Stripe Dashboard](https://dashboard.stripe.com/test/settings/billing/portal).
 
-```ts title=".env.server"
-STRIPE_CUSTOMER_PORTAL_URL=<your-test-customer-portal-link>
-```
+By default, OpenSaas generates a unique customer portal link for each user on the back end.
+If you'd rather provide a permanent link to the customer portal, activate and copy the `Customer portal link`.
```
infomiho:
```suggestion
If you'd rather provide a permanent link to the customer portal, activate it and copy the `Customer portal link`.
```

I'm not sure what are we activating and where. Maybe this is due to not having the UI in front of me, but let's make it a bit more clear in the text.

template/app/src/payment/stripe/checkoutUtils.ts
infomiho: @FranjoMindek you can use the `env` object here instead of going for `process.env`. The `env` object has the `http://localhost:3000/` as the default value: https://wasp.sh/docs/project/env-vars#server-env-vars

I just realized the docs don't say this:
SCR-20251104-jzxh

Made an issue for this: https://github.com/wasp-lang/wasp/issues/3294

template/app/src/payment/stripe/checkoutUtils.ts
```diff
@@ -1,64 +1,67 @@
...
+  const CLIENT_BASE_URL =
+    process.env.WASP_WEB_CLIENT_URL || "http://localhost:3000";
```
infomiho:
```suggestion
   env.WASP_WEB_CLIENT_URL;
```

template/app/src/payment/stripe/checkoutUtils.ts
```diff
@@ -1,64 +1,67 @@
...
+    },
+    // Stripe automatically creates invoices for subscriptions.
+    // For one-time payments, we must enable them manually.
+    // However, enabling invoices for subscriptions will throw an error.
+    invoice_creation:
+      mode === "payment"
+        ? {
+            enabled: true,
+          }
+        : undefined,
```
infomiho: I don't want us to overengineer this since it's a template, but maybe this logic could live in a small utility function e.g. `getInvoiceCreationConfig` or something like that. The comment could live above that function.

template/app/src/payment/stripe/webhookPayload.ts
```diff
@@ -1,118 +0,0 @@
-import { Stripe } from "stripe";
```
infomiho, line 1: I understand that this was redundant because Stripe checks the checksum. Do they check the payload in any way? Aren't we losing here the runtime guarantees.

And more importantly are we losing the uniformity between all the providers e.g. Stripe, LemonSqueezy, Polr?

I'd keep it in just for uniformity argument - the flow is always - get a payload, parse it, use it. What do you think?

template/app/src/payment/plans.ts
infomiho: Nice change!

template/app/src/payment/stripe/paymentProcessor.ts
```diff
@@ -35,26 +31,64 @@ export const stripePaymentProcessor: PaymentProcessor = {
...
+      },
     };
-    return { session };
   },
-  fetchCustomerPortalUrl: async (_args: FetchCustomerPortalUrlArgs) =>
-    requireNodeEnvVar("STRIPE_CUSTOMER_PORTAL_URL"),
+  fetchCustomerPortalUrl: async (args: FetchCustomerPortalUrlArgs) => {
+    // WASP_WEB_CLIENT_URL will be set up by Wasp when deploying to production: https://wasp.sh/docs/deploying
+    const CLIENT_BASE_URL =
+      process.env.WASP_WEB_CLIENT_URL || "http://localhost:3000";
```
infomiho:
```suggestion
    const CLIENT_BASE_URL = env.WASP_WEB_CLIENT_URL;
```

template/app/src/payment/stripe/paymentProcessor.ts
```diff
@@ -35,26 +31,64 @@ export const stripePaymentProcessor: PaymentProcessor = {
...
+
+    const user = await args.prismaUserDelegate.findUniqueOrThrow({
+      where: {
+        id: args.userId,
+      },
+      select: {
+        paymentProcessorUserId: true,
+      },
+    });
```
infomiho: This code here felt a little too low level, what about a helper e.g. `getUserPaymentProcessorId` or something like that.

template/app/src/payment/stripe/paymentProcessor.ts
```diff
@@ -35,26 +31,64 @@ export const stripePaymentProcessor: PaymentProcessor = {
...
+
+    if (!user.paymentProcessorUserId) {
+      return null;
+    }
+
+    const session = await stripeClient.billingPortal.sessions.create({
+      customer: user.paymentProcessorUserId,
+      return_url: `${CLIENT_BASE_URL}/account`,
+    });
```
infomiho: This could also be `getCustomerSessionFromUserId` or something like that. I think you'll figure it out 😄

template/app/src/payment/stripe/webhook.ts
```diff
@@ -7,59 +7,74 @@ import { emailSender } from "wasp/server/email";
...
-      return response.status(422).json({ error: err.message });
+      // We must return a 2XX status code, otherwise Stripe will keep retrying the event.
+      response.status(204).send();
+
+      // In development, it is likely that we will receive events that we are not handling.
+      // E.g. via the `stripe trigger` command.
+      // These can be ignored without any issues.
+      if (process.env.NODE_ENV === "production") {
+        throw err;
+      }
```
infomiho: Here we would be throwing an error _after_ we sent a response, can this cause some unexpected behavior?

template/app/src/payment/stripe/webhook.ts
```diff
@@ -71,210 +86,177 @@ export const stripeWebhook: PaymentsWebhook = async (
...
+  const invoicePaidAtDate = getInvoicePaidAtDate(invoice);
+  const paymentPlanId = getPaymentPlanIdByPriceId(getInvoicePriceId(invoice));
+
+  switch (paymentPlanId) {
+    case PaymentPlanId.Credits10:
+      await updateUserStripeOneTimePaymentDetails(
+        {
+          customerId,
+          datePaid: invoicePaidAtDate,
+          numOfCreditsPurchased: paymentPlans.credits10.effect.amount,
```
infomiho: Here we are getting the payment plan ID and then using the `paymentPlans` to get the amount. Can we update the helper to return plan objects e.g. `getPaymentPlanByPriceId` so we use `plan.effect.amount` - we can't make the mistake of the plan ID and the value going out of sync.

template/app/src/payment/stripe/webhook.ts
```diff
@@ -71,210 +86,177 @@ export const stripeWebhook: PaymentsWebhook = async (
...
-  let subscriptionStatus: SubscriptionStatus | undefined;
-  const priceId = extractPriceId(subscription.items);
-  const subscriptionPlan = getPlanIdByPriceId(priceId);
+): Promise<void> {
+  const subscription = event.data.object;
 
-  // There are other subscription statuses, such as `trialing` that we are not handling and simply ignore
-  // If you'd like to handle more statuses, you can add more cases above. Make sure to update the `SubscriptionStatus` type in `payment/plans.ts` as well.
+  // There are other subscription statuses, such as `trialing` that we are not handling.
+  let subscriptionStatus: SubscriptionStatus | undefined;
```
infomiho: Can we maybe create a helper fn for this e.g. `getSubscriptionStatus` so we can use a `const` still.

template/app/src/payment/stripe/webhook.ts
```diff
@@ -71,210 +86,177 @@ export const stripeWebhook: PaymentsWebhook = async (
...
-// Invoices are not created for one-time payments, so we handle them above.
-async function handleInvoicePaid(
-  invoice: InvoicePaidData,
-  prismaUserDelegate: PrismaClient["user"],
-) {
-  await saveActiveSubscription(invoice, prismaUserDelegate);
-}
+  const priceId = invoiceLineItems[0].pricing?.price_details?.price;
+  if (!priceId) {
+    throw new Error("Unable to extract price id from items");
```
infomiho, line 152: Nipick: Earlier we threw `HttpError` in this fn - do we want to use it for this throw as well?

template/app/src/payment/stripe/paymentDetails.ts
```diff
@@ -1,36 +1,57 @@
...
 import { PaymentPlanId } from "../plans";
 
-export const updateUserStripePaymentDetails = async (
+interface UpdateUserStripeOneTimePaymentDetails {
+  customerId: Stripe.Customer["id"];
+  datePaid: Date;
+  numOfCreditsPurchased: number;
+}
+
+export function updateUserStripeOneTimePaymentDetails(
```
infomiho: Do we want to add explicit return types to these fns? I know we want to do it in our codebase, but since this is a template we might want to more relaxed with this?

infomiho (commented):
Good stuff @FranjoMindek 

I think there is one "big" question to disucss - do we keep the Zod schemas for parsing the payload in the name of uniformity?

### Testing it locally

Running it locally with the `opensaas.sh` got me this error when running `wasp db migrate-dev`, maybe there is a migration missing?

```
[  Wasp  ] src/payment/stripe/paymentProcessor.ts(64,9): error TS2353: Object literal may only specify known properties, and 'paymentProcessorUserId' does not exist in type 'UserSelect<DefaultArgs>'.
[  Wasp  ] src/payment/stripe/paymentProcessor.ts(68,15): error TS2339: Property 'paymentProcessorUserId' does not exist on type '{ id: string; createdAt: Date; email: string | null; username: string | null; isAdmin: boolean; isMockUser: boolean; stripeId: string | null; subscriptionStatus: string | null; subscriptionPlan: string | null; datePaid: Date | null; credits: number; }'.
[  Wasp  ] src/payment/stripe/paymentProcessor.ts(73,22): error TS2339: Property 'paymentProcessorUserId' does not exist on type '{ id: string; createdAt: Date; email: string | null; username: string | null; isAdmin: boolean; isMockUser: boolean; stripeId: string | null; subscriptionStatus: string | null; subscriptionPlan: string | null; datePaid: Date | null; credits: number; }'.
```

template/app/src/payment/stripe/webhookPayload.ts
FranjoMindek: > I understand that this was redundant because Stripe checks the checksum. Do they check the payload in any way? Aren't we losing here the runtime guarantees.

The checksum is created from the payload data.

In theory, this check would fail only when Stripe malforms data on their end before creating the checksum. I think we can trust it here.

> And more importantly are we losing the uniformity between all the providers e.g. Stripe, LemonSqueezy, Polr?
> I'd keep it in just for uniformity argument - the flow is always - get a payload, parse it, use it. What do you think?

If LemonSqueezy is also doing checksum checks, which I think they do, and they do type their API properly, I think the way forward would be to also remove our own payload parsing from LemonSqueezy. 

Polar does check payload validity, and we don't do our own parsing in that PR. Though Polar also has nicer type API than Stripe.

While yes, the Stripe type API is not that great (it throws all data possibilities on you instead of narrowing them), I would rather keep types up to date with their API rather than inventing our own, which could be prone to becoming outdated.

template/app/src/payment/stripe/webhook.ts
FranjoMindek: Hm, I've already had dilemma before about this block of code.
I wasn't sure at that moment if I should crash the application or just log an error at that point if its production.

Now, when I think about it, I think that crashing the application was a bad choice (too aggressive). While this is something you want to fix, having extra webhook events is not a critical mistake. It's "recoverable".

I will change it to log the error in production, and log the warning otherwise.

template/app/src/payment/stripe/webhook.ts
FranjoMindek: With the `as const satisfies Record<PaymentPlanId, PaymentPlan>;` type change I did this should be enough:

image

I.e. instead of hardcoding `credits10` I can use `paymentPlanId` because switch case narrows it.
And because it is a const type, we get full type inferred correctly.

I like payment plan id approach more than payment plan object, because payment plan object itself hold no identifier.

template/app/src/payment/stripe/webhook.ts
FranjoMindek: I researched a bit and `HttpError` is Wasp error which we use when we want to provide additional data to the web client. Webhook events are not relayed to the web client in any way. So I would maybe remove all notion of `HttpError` as not to confuse people with it.

Does that make sense to you?

template/app/src/payment/stripe/paymentDetails.ts
FranjoMindek: Oh sure, I think I actually wasn't sure how to get proper prisma type here manually, so I skipped them.
Will add them.

template/app/src/payment/stripe/webhook.ts
FranjoMindek: Did it as `getOpenSaasSubscriptionStatus` since this is Stripe domain and `Stripe.SubscriptionStatus` exists.

opensaas-sh/blog/src/content/docs/guides/payments-integration.mdx
FranjoMindek: It's meant to continue on sentence above:

> You can set up your customer portal in your Stripe Dashboard...

So Stripe dashboard, customer portal page, there is a button to activate it.
But they changed wording on it a bit. Before it literally said `Customer portal link` IIRC.

Image

Image

FranjoMindek: I think I'll go with:

If you'd rather provide a permanent link to the customer portal, activate it and copy the `portal link`.

template/app/src/payment/stripe/checkoutUtils.ts
FranjoMindek: I just used `env.WASP_WEB_CLIENT_URL` directly. 👍🏻

template/app/src/payment/stripe/paymentProcessor.ts
FranjoMindek: I just used `env.WASP_WEB_CLIENT_URL` directly. 👍🏻

FranjoMindek:
> Running it locally with the opensaas.sh got me this error when running wasp db migrate-dev, maybe there is a migration missing?

It seems like I need to update the `opensaas-sh/app_diff`, because we don't have `paymentProcessorUserId` field in there, we have `stripeId` instead. Don't know how it disappeared from diffs where necessary.

edit: Should be fixed now.

opensaas-sh/blog/src/content/docs/guides/payments-integration.mdx
infomiho: One detail: make sure to match the case i.e. `Portal link` since that's the label in the Stripe UI.

template/app/src/payment/stripe/paymentProcessor.ts
FranjoMindek: I've extracted prisma logic into separate functions.

All of them are now part of `payment/stripe/user.ts` file. Since that file actually has all logic related to modifying/fetching user data in relation to stripe payment processing.

```ts
const paymentProcessorUserId = await fetchUserPaymentProcessorUserId(
  userId,
  prismaUserDelegate,
);
```

template/app/src/payment/stripe/paymentProcessor.ts
FranjoMindek: I've decided to keep this one since there is no real complexity that we can hide here.
I think this one is pretty straightforward.

opensaas-sh/app_diff/src/client/Main.css.diff
```diff
@@ -1,6 +1,6 @@
 --- template/app/src/client/Main.css
```
infomiho, line 1: How come we have these changes in this PR?

Should we maybe do a separate PR that only updates the diffs?

template/app/src/payment/paymentProcessor.ts
```diff
@@ -1,17 +1,18 @@
-import { PrismaClient } from "@prisma/client";
+import { PrismaClient, User } from "@prisma/client";
```
infomiho: It's more canonical to use `wasp/entities` for model imports

template/app/src/payment/stripe/user.ts
```diff
@@ -0,0 +1,92 @@
+import { PrismaClient, User } from "@prisma/client";
```
infomiho: Let's use `wasp/entities` here as well

template/app/src/payment/stripe/webhook.ts
infomiho: > when we want to provide additional data to the web client

There is one more use case, if you check how it's handled in our server code:

```js
if (err instanceof HttpError) {
  return res.status(err.statusCode).json({ message: err.message, data: err.data })
}
```

You'll notice it gives you control over the error code as well on the server. Also, it enables you to send some message in the error response vs. just 500 https://github.com/wasp-lang/wasp/blob/main/waspc/data/Generator/templates/server/src/app.js#L16

If you feel like we shouldn't send any extra info (status code, message) that's also fine 👍

template/app/src/payment/stripe/checkoutUtils.ts
```diff
@@ -1,64 +1,72 @@
-import type { StripeMode } from "./paymentProcessor";
-
+import { User } from "@prisma/client";
```
infomiho: `wasp/entities` here as well 😄

template/app/src/payment/stripe/webhook.ts
```diff
@@ -1,67 +1,85 @@
...
+  } catch (error) {
+    if (error instanceof UnhandledWebhookEventError) {
+      // In development, it is likely that we will receive events that we are not handling.
+      // E.g. via the `stripe trigger` command.
+      // While these can be ignored safely in development, it's good to be aware of them.
+      // For production we shouldn't have any extra webhook events.
+      if (process.env.NODE_ENV === "development") {
+        console.info("Unhandled Stripe webhook event in development: ", error);
+      } else if (process.env.NODE_ENV === "production") {
+        console.error("Unhandled Stripe webhook event in production: ", error);
```
infomiho, line 73: I'm afraid this might go unnoticed in the logs. But I guess, if you search for the event name - it will appear. How does this error look like in the logs now?

template/app/src/payment/stripe/webhook.ts
```diff
@@ -71,210 +89,182 @@ export const stripeWebhook: PaymentsWebhook = async (
...
-    planEffect: plan.effect,
-  });
-  return updateUserStripePaymentDetails(
-    { userStripeId, numOfCreditsPurchased, datePaid: new Date() },
-    prismaUserDelegate,
-  );
-}
+/**
+ * We only expect one line item, if your workflow expects more, you should change this function to handle them.
+ */
```
infomiho: Nitpick: This comment feels like a note from us to them so I'd maybe put it in line comments e.g.

```js
// We only expect one line item, if your workflow expects more, you should change this function to handle them.
```

This way it doesn't look like the function documentation.

template/app/src/payment/stripe/webhook.ts
```diff
@@ -71,210 +89,182 @@ export const stripeWebhook: PaymentsWebhook = async (
...
+    getSubscriptionPriceId(subscription),
+  );
+
+  const user = await updateUserSubscriptionDetails(
+    { customerId, paymentPlanId, subscriptionStatus },
     prismaUserDelegate,
   );
+
+  if (subscription.cancel_at_period_end) {
+    if (user.email) {
```
infomiho: Nitpick:
```suggestion
  if (subscription.cancel_at_period_end && user.email) {
```

template/app/src/payment/stripe/webhook.ts
```diff
@@ -71,210 +89,182 @@ export const stripeWebhook: PaymentsWebhook = async (
...
 
-async function handleCustomerSubscriptionUpdated(
-  subscription: SubscriptionUpdatedData,
-  prismaUserDelegate: PrismaClient["user"],
-) {
-  const userStripeId = subscription.customer;
+function getOpenSaasSubscriptionStatus(
+  subscription: Stripe.Subscription,
+): SubscriptionStatus | undefined {
   let subscriptionStatus: SubscriptionStatus | undefined;
```
infomiho: Why not return right away vs. using a `let` and returning at the end?

template/app/src/payment/stripe/webhookPayload.ts
infomiho: > If LemonSqueezy is also doing checksum checks, which I think they do, and they do type their API properly, I think the way forward would be to also remove our own payload parsing from LemonSqueezy.

If this is the case for both LemonSqueezy and Polr - then I'm okay with not parsing the payload with Zod 👍  

Should we then update our LemonSqueezy not to use Zod validation for payload as well? (Separate PR of course)

RAW

opensaas-sh/app_diff/src/client/Main.css.diff
FranjoMindek: Yep, this often happens.
We should have a CI workflow for this.

template/app/src/payment/stripe/webhook.ts
FranjoMindek: Currently it logs the error itself, so it's pretty noticeable.
Maybe even tone it down for development?

image

template/app/src/payment/stripe/webhook.ts
FranjoMindek: I just extracted the logic out, didn't improve it, my bad.

template/app/src/payment/stripe/webhook.ts
FranjoMindek: I guess that would be valid for self-documentation reasons.

Stripe only recognizes if its 2XX status code or anything else.
400 and 500 make no difference to it.

I'm fine with just `Error` here, if you are also, you can resolve, otherwise comment.

template/app/src/payment/stripe/webhook.ts
infomiho: Oh this looks pretty noticeable yeah!

I think it's fine if it stays like this in development as far as I'm concerned 👍 

RAW

opensaas-sh/app_diff/src/client/Main.css.diff
infomiho: What kind of workflow were you thinking? It is something we should encode in an improvement issue?

template/app/src/payment/stripe/checkoutUtils.ts
```diff
@@ -1,64 +1,72 @@
...
+    email: userEmail,
+  });
 
-export async function fetchStripeCustomer(customerEmail: string) {
-  let customer: Stripe.Customer;
-  try {
-    const stripeCustomers = await stripe.customers.list({
-      email: customerEmail,
+  if (stripeCustomers.data.length === 0) {
+    console.log("Creating a new Stripe customer");
```
infomiho: I just noticed these logs (I know they are old code) - do we find them useful? They feel a bit like debug logs vs. logs we want to keep. 

RAW

template/app/src/payment/stripe/webhook.ts
infomiho: Hey, it looks like there were two places to make this change? I see old code here but the change in this commit: https://github.com/wasp-lang/open-saas/pull/505/commits/99f01adb5b28ecf581d04a6da0d0c211bd8eaa30

infomiho (approved):
Good to go! 

Left some small comments.

template/app/src/payment/stripe/webhook.ts
FranjoMindek: My bad, fixed the second one now.

template/app/src/payment/stripe/webhookPayload.ts
FranjoMindek: I'm up for refreshing LemonSqueezy too.
Post Polar though.

opensaas-sh/app_diff/src/client/Main.css.diff
FranjoMindek: A workflow which would check if generating a diff makes changes to the working directory. If it does it fails. Just to make sure your diffs are up to date with your template.

template/app/src/payment/stripe/checkoutUtils.ts
FranjoMindek: I'm up for removing these. I've kept them for historical reasons.

I would keep logs only if we would introduce an actual logging library, and then I would do it with appropriate log levels. 

I don't really like them in current approach.

However, I wouldn't do that as part of this PR.

template/app/src/payment/stripe/webhookPayload.ts
infomiho: Ok let's create issues for the things we agreed on 👍

opensaas-sh/app_diff/src/client/Main.css.diff
infomiho: Let's create an issue if it makes sense 👍

Issue #503, opened 2025-09-15
User credits not displayed after cancelling subscription


Genyus:
If a user purchases some credits, then purchases a subscription, but then that sub is revoked/cancelled, while the previously purchased credits remain on their account, the user will never be able to see how many they have due to the current logic that [checks for the subscription status, plan ID and payment date](https://github.com/wasp-lang/open-saas/blob/fabbfef0e93aaa35f1e9ae48f6c83f5e2f240791/template/app/src/user/AccountPage.tsx#L78-L95).

PR #506, opened 2025-09-17
Account page improvements

FranjoMindek:
Always display both subscription plan and credits information. 
Fixes https://github.com/wasp-lang/open-saas/issues/503

Always display customer portal link button if possible.

Rename the "buy more/upgrade" button to "buy more credits" and move it to credits row. Display it only if no subscription.
Why? If we want to upgrade subscription we can always use the "manage payment details" button.
Only thing we need to navigate to pricing page for is to buy more credits.

### DIFF
--- a/template/app/src/user/AccountPage.tsx
+++ b/template/app/src/user/AccountPage.tsx
@@ -45,14 +45,21 @@ export default function AccountPage({ user }: { user: User }) {
             <div className='py-4 px-6'>
               <div className='grid grid-cols-1 sm:grid-cols-3 sm:gap-4'>
                 <dt className='text-sm font-medium text-muted-foreground'>Your Plan</dt>
-                <UserCurrentSubscriptionPlan user={user} />
+                <UserCurrentSubscriptionPlan
+                  subscriptionPlan={user.subscriptionPlan}
+                  subscriptionStatus={user.subscriptionStatus}
+                  datePaid={user.datePaid}
+                />
               </div>
             </div>
             <Separator />
             <div className='py-4 px-6'>
               <div className='grid grid-cols-1 sm:grid-cols-3 sm:gap-4'>
                 <dt className='text-sm font-medium text-muted-foreground'>Credits</dt>
-                <dd className='mt-1 text-sm text-foreground sm:col-span-2 sm:mt-0'>{user.credits}</dd>
+                <dd className='mt-1 text-sm text-foreground sm:col-span-1 sm:mt-0'>{user.credits}</dd>{' '}
+                <div className='mt-4 sm:mt-0 ml-auto'>
+                  <BuyMoreButton subscriptionStatus={user.subscriptionStatus} />
+                </div>
               </div>
             </div>
             <Separator />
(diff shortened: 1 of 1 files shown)


template/app/src/user/AccountPage.tsx
```diff
@@ -62,119 +76,88 @@ export default function AccountPage({ user }: { user: User }) {
...
-      >
-        Manage Subscription
-      </Button>
-    </div>
+    <WaspRouterLink
+      to={routes.PricingPageRoute.to}
+      className='font-medium text-sm text-primary hover:text-primary/80 transition-colors duration-200'
+    >
+      <Button variant='link'>Buy More Credits</Button>
+    </WaspRouterLink>
```
infomiho, line 193: Don't we have some sort of `asChild` or `as` attribute on `Link` so it renders as a button?

template/app/src/user/AccountPage.tsx
```diff
@@ -40,15 +45,24 @@ export default function AccountPage({ user }: { user: User }) {
...
+                  subscriptionStatus={user.subscriptionStatus}
                   datePaid={user.datePaid}
-                  credits={user.credits}
                 />
               </div>
             </div>
             <Separator />
+            <div className='py-4 px-6'>
+              <div className='grid grid-cols-1 sm:grid-cols-3 sm:gap-4'>
+                <dt className='text-sm font-medium text-muted-foreground'>Credits</dt>
```
infomiho: Should `dt` be used in `dl`? https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dt

FranjoMindek:
@infomiho this is not urgent, but you did self-assign.
Just commenting in case you forgot you self-assigned.

template/app/src/user/AccountPage.tsx
```diff
@@ -59,17 +60,28 @@ export default function AccountPage({ user }: { user: User }) {
...
               </div>
             </div>
             <Separator />
+            <div className="px-6 py-4">
+              <div className="grid grid-cols-1 sm:grid-cols-3 sm:gap-4">
+                <dt className="text-muted-foreground text-sm font-medium">
+                  Credits
+                </dt>
+                <dd className="text-foreground mt-1 text-sm sm:col-span-1 sm:mt-0">
+                  {user.credits + " credits"}
```
infomiho:
```suggestion
                  {user.credits} credits
```

infomiho (approved):
Smaller comments - sorry it took so long to review!

template/app/src/user/AccountPage.tsx
FranjoMindek: This screws up the `<dd>` rendering. It makes it have 2 children instead of one.
Which screws up `locator` text search in `playwright`. 
Actually wanted to have it that way. 😞

template/app/src/user/AccountPage.tsx
FranjoMindek: Not as far as I can tell.

template/app/src/user/AccountPage.tsx
FranjoMindek: This was old code.
From what I can see we use "description list" but without `<dl>` to emulate what looks like a table.
Really weird.

If we want a table we should use table. I think it might be best to turn all of these into `<div>` for now.
Then later we can create a proper table.

template/app/src/user/AccountPage.tsx
FranjoMindek: Actually works now that we no longer have `<dd>`

PR #508, opened 2025-09-18
Add root prettier configuration with the same settings as our `wasp` repo 

FranjoMindek:
## Description

This is NOT prettier configuration we are shipping to users with template.
This is repository root configuration to keep the whole repository formatted.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -5,5 +5,9 @@ dist/
 # OpenSaaS app diff folder.
 app_diff/
 
+# Formatting `.md` and `.mdx` can introduce logical changes.
+**/blog/**/*.md
+**/blog/**/*.mdx 
+
 # Ignore minified JS files in the public folder.
 **/public/**/*.js
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,9 @@
+# Build outputs
+.wasp/
+dist/
+
+# OpenSaaS app diff folder.
+app_diff/
+
+# Ignore minified JS files in the public folder
+**/public/**/*.js
--- a/.prettierignore
+++ b/.prettierignore
@@ -5,7 +5,7 @@ dist/
 # OpenSaaS app diff folder.
 app_diff/
 
-# Formatting `.md` and `.mdx` can introduce logical changes.
+# Formatting `.md` and `.mdx` files can introduce logical changes.
 **/blog/**/*.md
 **/blog/**/*.mdx 
 
--- a/.prettierignore
+++ b/.prettierignore
@@ -7,7 +7,7 @@ app_diff/
 
 # Formatting `.md` and `.mdx` files can introduce logical changes.
 **/blog/**/*.md
-**/blog/**/*.mdx 
+**/blog/**/*.mdx
 
 # Ignore minified JS files in the public folder.
 **/public/**/*.js
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,9 +1,9 @@
-# Build outputs
+# Build outputs.
 .wasp/
 dist/
 
 # OpenSaaS app diff folder.
 app_diff/
 
-# Ignore minified JS files in the public folder
+# Ignore minified JS files in the public folder.
 **/public/**/*.js
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "prettier-plugin-organize-imports": "4.1.0",
     "prettier-plugin-pkg": "0.19.1",
     "prettier-plugin-sh": "0.17.4",
-    "prettier-plugin-tailwindcss": "^0.6.13",
+    "prettier-plugin-tailwindcss": "0.6.13",
     "typescript": "~5.8.3"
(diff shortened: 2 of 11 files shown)


template/blog/.gitignore
```diff
@@ -1,24 +1,18 @@
...
 
-# environment variables
+# Ignore all dotenv files by default to prevent accidentally committing any secrets.
+# To include specific dotenv files, use the `!` operator or adjust these rules.
 .env
-.env.production
+.env.*
 
-# macOS-specific files
-.DS_Store
```
FranjoMindek, line 24: Moved to root.

cprecioso: We're not `.gitignore`ing the .DS_Store anymore? I think it might be the case that some of our users don't have their global git ignore configured.

template/blog/.gitignore
```diff
@@ -1,24 +1,18 @@
...
+dist/
 
+# Deployment
+.netlify/
 
-# environment variables
+# Ignore all dotenv files by default to prevent accidentally committing any secrets.
+# To include specific dotenv files, use the `!` operator or adjust these rules.
 .env
-.env.production
```
cprecioso: I'd like someone to chime in wrt whether this was done on purpose

template/blog/.gitignore
```diff
@@ -1,24 +1,18 @@
...
-
-# dependencies
+# Dependencies
 node_modules/
 
-# logs
-npm-debug.log*
-yarn-debug.log*
-yarn-error.log*
-pnpm-debug.log*
```
cprecioso, line 16: Why remove this?

cprecioso (commented):
As for the `prettier` changes: LGTM, but I'd like to already see a PR on top of this one with the changes applied, so we can inspect it a bit, and see if we need to ignore more files.

The `gitignore` changes: I see your changes and mostly would do them myself for a project of mine. But I'm not sure if we're targeting people that might not be completely comfortable with gitignores and it is useful to have some boilerplate there.

template/blog/.gitignore
FranjoMindek: I think the weird part was that we ignored it on `blog` but not on `e2e-tests` and `app`.
So we ignored it for users partially.

Ideally we would either add `template/.gitignore` and do it properly, or let mac users add it themselves (assume nothing about user).

template/blog/.gitignore
FranjoMindek: I would ask Why did we have them in the first place?
Also why only on `blog`?

I think that this is really niche, really not necessary.
People who generate those certainly wont commit them.

I think they were vibecoded in the first place.

template/blog/.gitignore
FranjoMindek: I assume Vinny added this when initially building `open-saas`.
However, we don't use any `.env` variables in `blog`. We could straight up delete these git rules.

template/blog/.gitignore
cprecioso: > Ideally we would either add template/.gitignore and do it properly

That sounds right

FranjoMindek:
Added prettier format write PR here: https://github.com/wasp-lang/open-saas/pull/509

template/blog/.gitignore
FranjoMindek: I've moved all env rules to `template/.gitignore`.

package.json
```diff
@@ -0,0 +1,15 @@
...
+  "scripts": {
+    "prettier:check": "prettier --ignore-unknown --check --config .prettierrc .",
+    "prettier:format": "prettier --ignore-unknown --write --config .prettierrc ."
+  },
+  "devDependencies": {
+    "prettier": "3.5.3",
+    "prettier-plugin-organize-imports": "4.1.0",
+    "prettier-plugin-pkg": "0.19.1",
+    "prettier-plugin-sh": "0.17.4",
+    "prettier-plugin-tailwindcss": "^0.6.13",
```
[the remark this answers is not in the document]
infomiho: Good catch, I guess we'd want to be consistent?

.prettierignore
```diff
@@ -0,0 +1,13 @@
+# Build outputs.
+.wasp/
+dist/
+
+# OpenSaaS app diff folder.
+app_diff/
+
+# Formatting `.md` and `.mdx` files can introduce logical changes.
+**/blog/**/*.md
+**/blog/**/*.mdx
```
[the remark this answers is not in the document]
infomiho: Simple enough

infomiho (approved):
I'd suggest addressing the Copilots comments

package.json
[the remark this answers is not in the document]
FranjoMindek: This is copy-paste from wasp repo. Means we should have the bug there too.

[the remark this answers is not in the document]
cprecioso: Yep, prettier and its plugins are usually recommended to have an exact version and not a caret

PR #513, opened 2025-09-24
Git ignore `opensaas-sh/app/` directory

FranjoMindek:
## Description

People who work on Open SaaS are often pained by `opensaas-sh/app/` folder being tracked.

By forcing the `opensaas-sh/app/` to be a separate git repository (via `git init`), we can:
1) Git ignore it
2) Have rest of diffing logic work (`git ls-files`) because there are no changes to git ignored files from the perspective of `opensaas-sh/app/`.

### DIFF
--- a/opensaas-sh/.gitignore
+++ b/opensaas-sh/.gitignore
@@ -1,3 +1,1 @@
-# We can't ignore `app/` because it messes up our patch/diff procedure (check
-# the README for more info on this)
-# app/
+app/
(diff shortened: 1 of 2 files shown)


infomiho:
Do you know why I get new diffs when I run this:
```
./tools/patch.sh
./tools/diff.sh
```
locally?

FranjoMindek:
Hm, I don't get them locally.

FranjoMindek:
Did you clean up the `./app` directory beforehand?
~~If you use `patch.sh` on top of already generated app, you modify it.~~

infomiho:
I didn't clean up the old `app`, after doing that - this no longer happens 👍

infomiho (approved):

PR #514, opened 2025-09-24
Add workflow to protect from uncomitted LLM text files changes

FranjoMindek:
Detects if there are any missing changes to LLM text files if we made changes to `opensaas-sh` docs.

Here is how it looks like when it:
1) fails: https://github.com/wasp-lang/open-saas/actions/runs/17981743175/job/51149343704?pr=514
2) passes: https://github.com/wasp-lang/open-saas/actions/runs/17981869647/job/51149798437?pr=514

Vibecoded but edited manually to fix echo + quiet mode diff.

### DIFF
--- a/opensaas-sh/blog/public/llms-full.txt
+++ b/opensaas-sh/blog/public/llms-full.txt
@@ -807,8 +807,6 @@ Setting up your app's authentication is easy with Wasp. In fact, it's already se
   },
 ```
 
-asd
-
 The great part is, by defining your auth config in the `main.wasp` file, Wasp manages most of the Auth process for you, including the auth-related database entities for user credentials and sessions, as well as auto-generated client components for your app on the fly (aka AuthUI -- you can see them in use in the `src/auth` folder).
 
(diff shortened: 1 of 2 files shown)


infomiho:
Hey we did this a bit differently in the https://github.com/wasp-lang/wasp repo, we don't keep the `llms.txt` in version control but generate them when we build the docs.
- Running on [Cloudflare](https://dash.cloudflare.com/75c37602a6b75c790961219411cb1b92/pages/view/wasp-docs/49f4fa06-c04e-434f-ab83-4656a6c440d6) on deploy
- The `build` script https://github.com/wasp-lang/wasp/blob/main/web/package.json#L10
- Ignoring `llms.txt` in `.gitignore` https://github.com/wasp-lang/wasp/blob/main/web/.gitignore#L23

FranjoMindek:
Oh didn't know we did that already on wasp.
I will modify the current approach to be same as above then. Much better than duplicating same changes in diffs.
Thanks.

Issue #516, opened 2025-09-26
Reduce the `template` size


infomiho:
Unpacked the `template` is 14 MB as it sits now. The biggest offenders are:
```
3.8M	./app/src/client/static/open-saas-banner-dark.png
2.2M	./app/src/client/static/open-saas-banner-light.png
1.8M	./blog/public/CRAIG_ROCK.png
336K	./blog/package-lock.json
220K	./app/src/client/static/assets/aiready.webp
220K	./app/src/client/static/assets/aiready-dark.webp
196K	./app/src/client/static/assets/admin.webp
172K	./app/src/client/static/assets/payments.webp
168K	./app/public/favicon.ico
148K	./blog/public/banner-images/default-banner.webp
```

We should really switch the banners with their SVG equivalent or get rid of the completely. 

We also should consider look into removing the custom font since it sits at around 3 MBs.

cprecioso:
Cut it in half (to 7MB) through #518 and #521 !

infomiho:
@cprecioso what would be the next steps to close this?

cprecioso:
Check on the font thing you mentioned last

PR #518, opened 2025-09-29
Ran lossless optimization on assets for template

cprecioso:
- Part of https://github.com/wasp-lang/open-saas/issues/516

Just a quick win. Ran [ImageOptim](https://imageoptim.com/mac) on all the `template` images. I used lossless settings, which means the output is pixel-for-pixel equal but the binary date was re-encoded/re-compressed, favoring the smallest possible output size.

image


template/app/src/client/icons/icon-copy-alt.svg
```diff
@@ -1,6 +1 @@
...
...3.org/2000/svg"><g fill="#fff"><path d="m17.6875 4.125-3.2813-3.25c-.2187-.21875-.5312-.34375-.8437-.34375h-5.6875c-.90625 0-1.65625.75-1.65625 1.65625v11.4062c0 .9063.75 1.6563 1.65625 1.6563h8.5c.9062 0 1.6562-.75 1.6562-1.6563v-8.62495c0-.3125-.125-.625-.3437-.84375zm-3.2188-1.1875 1.1875 1.1875h-1.1875zm1.9063 10.9062h-8.5c-.125 0-.25-.125-.25-.25v-11.4062c0-.125.125-.25.25-.25h5.1875v2.875c0 .375.3125.71875.7187.71875h2.8438v8.09375c0 .125-.125.2187-.25.2187z"/><path d="m13.7812 7.03125h-4.12495c-.375 0-.71875.3125-.71875.71875s.3125.71875.71875.71875h4.12495c.375 0 .7188-.3125.7188-.71875s-.3438-.71875-.7188-.71875z"/><path d="m13.7812 9.65625h-4.12495c-.375 0-.71875.3125-.71875.71875 0 .375.3125.7187.71875.7187h4.12495c.375 0 .7188-.3124.7188-.7187-.0313-.40625-.3438-.71875-.7188-.71875z"/><path d="m13.0625 16.25c-.375 0-.7187.3125-.7187.7187v.8438c0 .125-.125.25-.25.25h-8.4688c-.125 0-.25-.125-.25-.25v-11.4375c0-.125.125-.25.25-.25h1.0625c.375 0 .71875-.3125.71875-.71875s-.3125-.71875-.71875-.71875h-1.0625c-.90625 0-1.65625.75-1.65625 1.65625v11.46875c0 .9063.75 1.6562 1.65625 1.6562h8.5c.9063 0 1.6562-.7499 1.6562-1.6562v-.8438c0-.4062-.3125-.7187-.7187-.7187z"/></g></svg>
```
FranjoMindek, line 1: I'm not sure if prettier formats `.svg` files.
But here we are missing EOL. Should we miss it on `.svg` files? Try running prettier in root.
We still don't have format check workflow here.

FranjoMindek (approved):
I'm sure you've researched the tool enough so we can be sure there isn't any quality downgrade.

I see some mention of ultra HDR being filtered out, but I don't think we have such use case. Make sure you are 100% safe with running this.
https://www.reddit.com/r/macapps/comments/1exkgs3/comment/mao7tz4

Down the line we will probably want `.webp`. (Google loves it, good for SEO)

Just check how prettier interacts with the one svg file we have.

template/app/src/client/icons/icon-copy-alt.svg
cprecioso: Prettier does not run on .svg files without plugins; and here the EOL is missing on purpose. Even though .svgs are textual content, they are not used as source code here, but as regular assets; that's why they are neither formatted nor have an EOL, as they are optimized for size.

cprecioso: Also, checked if any images have HDR and they don't seem to.

PR #520, opened 2025-10-01
Remove note about ignoring `app`

cprecioso:


FranjoMindek (approved):
My bad, forgot we had that one.

PR #525, opened 2025-10-02
Add `*.copy` support to `dope.sh`

cprecioso:
- Part of https://github.com/wasp-lang/open-saas/issues/516
- Part of #526

First step for #526, this PR adds support for binary files in the `app_diff`.

- In `diff.sh`, before generating a patchfile, we first check if the file is binary. If so, we just copy it to the `app_diff` with a `.copy` suffix
- In `patch.sh`, after applying all the `.diff` files, we then apply all the `.copy` files by just copying them to `app`, without the suffix

This functionality is not used in this PR, but in the next one #526

> [!NOTE]
> Made by 🤖 Claude and checked by 👨🏼 me

### DIFF
--- a/opensaas-sh/tools/dope.sh
+++ b/opensaas-sh/tools/dope.sh
@@ -40,22 +40,11 @@ DIFF_DIR="${DERIVED_DIR}_diff"
 DIFF_DIR_DELETIONS="${DIFF_DIR}/deletions"
 
 file_is_binary() {
-  local file=$1
-  if file --mime "$file" | grep -q "charset=binary"; then
-    return 0
-  else
-    return 1
-  fi
+  file --mime "$1" | grep -q "charset=binary"
 }
 
 files_are_equal() {
-  local file1=$1
-  local file2=$2
-  if cmp -s "$file1" "$file2"; then
-    return 0
-  else
-    return 1
-  fi
+  cmp -s "$1" "$2"
 }
 
 # Based on base dir and derived dir, creates a diff dir that contains the diff between the two dirs.
@@ -88,11 +77,9 @@ recreate_diff_dir() {
     mkdir -p "${DIFF_DIR}/$(dirname "${filepath}")"
 
     if file_is_binary "${derivedFilepath}"; then
-      # Generate a .copy file.
       cp "${derivedFilepath}" "${DIFF_DIR}/${filepath}.copy"
       echo "Generated ${DIFF_DIR}/${filepath}.copy"
     else
-      # Generate a .diff file.
       diff -Nu --label "${baseFilepath}" --label "${derivedFilepath}" "${filepathToBeUsedAsBase}" "${derivedFilepath}" > "${DIFF_DIR}/${filepath}.diff"
       echo "Generated ${DIFF_DIR}/${filepath}.diff"
     fi


opensaas-sh/tools/dope.sh
```diff
@@ -91,7 +110,7 @@ recreate_derived_dir() {
 
   # Copy all the source files from the base dir over to the derived dir.
   while IFS= read -r filepath; do
-    mkdir -p "${DERIVED_DIR}/$(dirname ${filepath})"
+    mkdir -p "${DERIVED_DIR}/$(dirname "${filepath}")"
```
cprecioso, line 108: Fixing a shellcheck warning

opensaas-sh/tools/dope.sh
```diff
@@ -100,7 +119,7 @@ recreate_derived_dir() {
   local num_patches_failed=0
   while IFS= read -r diff_filepath; do
     local derived_filepath
-    derived_filepath="${diff_filepath#${DIFF_DIR}/}"
+    derived_filepath="${diff_filepath#"${DIFF_DIR}"/}"
```
cprecioso, line 117: Fixing a shellcheck warning

opensaas-sh/tools/dope.sh
```diff
@@ -62,12 +62,31 @@ recreate_diff_dir() {
...
-    if [ $? -eq 1 ]; then
-      mkdir -p "${DIFF_DIR}/$(dirname "${filepath}")"
-      echo "${DIFF_OUTPUT}" > "${DIFF_DIR}/${filepath}.diff"
-      echo "Generated ${DIFF_DIR}/${filepath}.diff"
+    # Check if the file is binary
+    if file --mime "${derivedFilepath}" | grep -q "charset=binary"; then
+      # For binary files, check if they differ from the base file
+      local files_differ=1
+      if [ -f "${filepathToBeUsedAsBase}" ] && [ "${filepathToBeUsedAsBase}" != "/dev/null" ]; then
+        if cmp -s "${filepathToBeUsedAsBase}" "${derivedFilepath}"; then
```
FranjoMindek: I'm not a Linux commands expert.
Why are we not using hash based comparison command (using e.g. `md5`?).
We even use the `-s` flag.

What I'm trying to say is that we are not concerned with the contents (and actual binary differences) of the files. Only if they are same.

FranjoMindek (commented):
I'm gonna be a weak reviewer on here as Martin loves these.

The logic seems fine.
We only create diffs on difference i.e. no unnecessary duplication.

Nice initiative for making the template cleaner.

opensaas-sh/tools/dope.sh
cprecioso: > Why are we not using hash based comparison command?

Mmmm I wouldn't hash anything here, then we'd be reading and hashing both whole files. `cmp` will just start reading the files and exit whenever the first difference is found. Seems like our usecase is exactly the intended one.

FranjoMindek: Yes don't know too much here.
I would assume hashes use bulk instructions.
Both should be fine.

opensaas-sh/tools/dope.sh
```diff
@@ -62,12 +62,31 @@ recreate_diff_dir() {
       filepathToBeUsedAsBase="/dev/null"
     fi
 
-    local DIFF_OUTPUT
-    DIFF_OUTPUT=$(diff -Nu --label "${baseFilepath}" --label "${derivedFilepath}" "${filepathToBeUsedAsBase}" "${derivedFilepath}")
-    if [ $? -eq 1 ]; then
-      mkdir -p "${DIFF_DIR}/$(dirname "${filepath}")"
-      echo "${DIFF_OUTPUT}" > "${DIFF_DIR}/${filepath}.diff"
-      echo "Generated ${DIFF_DIR}/${filepath}.diff"
+    # Check if the file is binary
```
Martinsos:
```suggestion
    # Check if the file is binary.
```

opensaas-sh/tools/dope.sh
```diff
@@ -62,12 +62,31 @@ recreate_diff_dir() {
...
+    if file --mime "${derivedFilepath}" | grep -q "charset=binary"; then
+      # For binary files, check if they differ from the base file
```
Martinsos:
```suggestion
      # For binary files, check if they differ from the base file.
```

opensaas-sh/tools/dope.sh
```diff
@@ -62,12 +62,31 @@ recreate_diff_dir() {
...
     fi
 
-    local DIFF_OUTPUT
-    DIFF_OUTPUT=$(diff -Nu --label "${baseFilepath}" --label "${derivedFilepath}" "${filepathToBeUsedAsBase}" "${derivedFilepath}")
-    if [ $? -eq 1 ]; then
-      mkdir -p "${DIFF_DIR}/$(dirname "${filepath}")"
-      echo "${DIFF_OUTPUT}" > "${DIFF_DIR}/${filepath}.diff"
-      echo "Generated ${DIFF_DIR}/${filepath}.diff"
+    # Check if the file is binary
+    if file --mime "${derivedFilepath}" | grep -q "charset=binary"; then
```
Martinsos: Maybe extract this as a helper function? You won't need the comment above then + code is nicer potentially.

opensaas-sh/tools/dope.sh
Martinsos: But this comment also sounds redundant. We know it is for binary files, we are in the IF that checked for it. And check if t hey differ from the base file, that is obvious from the code below.

opensaas-sh/tools/dope.sh
```diff
@@ -62,12 +62,31 @@ recreate_diff_dir() {
...
+      # For binary files, check if they differ from the base file
+      local files_differ=1
+      if [ -f "${filepathToBeUsedAsBase}" ] && [ "${filepathToBeUsedAsBase}" != "/dev/null" ]; then
```
Martinsos: Why do you have to check for /dev/null, I forgot a bit what filepathToBeuSedAsBase is exactly, but I would assume there is no chance really of it being /dev/null (or is there)?

opensaas-sh/tools/dope.sh
```diff
@@ -62,12 +62,31 @@ recreate_diff_dir() {
...
-    local DIFF_OUTPUT
-    DIFF_OUTPUT=$(diff -Nu --label "${baseFilepath}" --label "${derivedFilepath}" "${filepathToBeUsedAsBase}" "${derivedFilepath}")
-    if [ $? -eq 1 ]; then
-      mkdir -p "${DIFF_DIR}/$(dirname "${filepath}")"
-      echo "${DIFF_OUTPUT}" > "${DIFF_DIR}/${filepath}.diff"
-      echo "Generated ${DIFF_DIR}/${filepath}.diff"
+    # Check if the file is binary
+    if file --mime "${derivedFilepath}" | grep -q "charset=binary"; then
+      # For binary files, check if they differ from the base file
+      local files_differ=1
```
Martinsos: Maybe rename it to `is_different_than_base_file`. Or even `is_binary_different_than_base_file`.

opensaas-sh/tools/dope.sh
```diff
@@ -62,12 +62,31 @@ recreate_diff_dir() {
...
+      if [ -f "${filepathToBeUsedAsBase}" ] && [ "${filepathToBeUsedAsBase}" != "/dev/null" ]; then
+        if cmp -s "${filepathToBeUsedAsBase}" "${derivedFilepath}"; then
+          files_differ=0
+        fi
+      fi
+
+      if [ ${files_differ} -eq 1 ]; then
+        # Only copy if files differ or base file doesn't exist
```
Martinsos: In this comment you are repeating the logic from above. Would be great if you didn't have to.
Maybe you can just remove this logic and it is clear enough.

I wonder why you had the need to say "if files differ or base file doesn't exist". That tells me the name `files_differ` is not correct, in the sense it doesn't capture enough, it doesn't capture this second part of the condition in its name.

Maybe we can turn it around and say "we are doing the copying if files are not the same". If there is no base file, it is kind of logical they are not the same, so it holds. So we could rename `files_differ` to `files_are_same` or `same_as_base_file` or something (and adjust logic a bit if needed) and then you can drop this comment.

opensaas-sh/tools/dope.sh
```diff
@@ -62,12 +62,31 @@ recreate_diff_dir() {
...
+        mkdir -p "${DIFF_DIR}/$(dirname "${filepath}")"
+        cp "${derivedFilepath}" "${DIFF_DIR}/${filepath}.copy"
+        echo "Generated ${DIFF_DIR}/${filepath}.copy (binary)"
+      fi
+    else
+      # For text files, generate a diff
```
Martinsos:
```suggestion
      # It is not a binary, so let's generate a diff.
```
I am not sure you even need this comment though.

opensaas-sh/tools/dope.sh
Martinsos: Interesting, this way of checking if it is binary! I don't konw how to do it, but this looks easy enough which is nice. Btw binary -> what exactly clasifies as binary? Could it be there are some binaries that won't pass this check? Are images binaries? Is binary always the opposite of text, or is there something in between?

cprecioso: It seems this `file --mime` incantation is the generally recommended way to check for binary-ness online.

Could it be there are some binaries that won't pass this check? -> I guess there could be some edge case somewhere, but this is used in loads of places and is a well-trodden path. I can't imagine us using such a weird file format in the opensaas template that we'd trip this check.

What qualifies as binary? -> It seems to be a bit loose among each tool, but in this usecase it's mostly any stream that consists of (or might consist of) byte strings that are not well-formed text codepoints. So yeah, opposite of textual content.

The NUL byte in particular seems to cause warnings and errors (both for bash pipes, and for the diff/patch utilities). The font files contain that byte, that's the main motivation for this change.

opensaas-sh/tools/dope.sh
```diff
@@ -62,11 +81,19 @@ recreate_diff_dir() {
       filepathToBeUsedAsBase="/dev/null"
     fi
 
-    local DIFF_OUTPUT
-    DIFF_OUTPUT=$(diff -Nu --label "${baseFilepath}" --label "${derivedFilepath}" "${filepathToBeUsedAsBase}" "${derivedFilepath}")
-    if [ $? -eq 1 ]; then
```
cprecioso, line 67: This thing of storing the output and checking the error code was done to skip files that are equal. Now we explicitly check for that before, so we can write the diff file directly, and the code is much simpler.

opensaas-sh/tools/dope.sh
```diff
@@ -62,11 +81,19 @@ recreate_diff_dir() {
...
-      mkdir -p "${DIFF_DIR}/$(dirname "${filepath}")"
-      echo "${DIFF_OUTPUT}" > "${DIFF_DIR}/${filepath}.diff"
+    if files_are_equal "${filepathToBeUsedAsBase}" "${derivedFilepath}"; then
+      continue
+    fi
+
+    mkdir -p "${DIFF_DIR}/$(dirname "${filepath}")"
+
+    if file_is_binary "${derivedFilepath}"; then
+      # Generate a .copy file.
```
cprecioso: I know that this comment (and the one on the `else`) is a bit redundant, but the shell syntax is verbose and a bit foreign, so I want to demarcate these sections in case somebody in the future is quickly scanning the code. Will remove if you think is unnecessary

cprecioso:
@Martinsos check again pls.
Refactored it a bit, extracted to named functions so that I could remove most of the comments (left two, as kind of headers)

opensaas-sh/tools/dope.sh
Martinsos: Ok, makes sense! raw

opensaas-sh/tools/dope.sh
Martinsos: In this case I agree. Bash is such a language that unfortunately it is not so easy to make things always readable without going overly out of the way, so having comments that are "redundant" can indeed be valuable. Also, it is ok to assume we all know bash moderately / not very well, so extra comments also help there. So it makes sense to write more comments in Bash than in let's say TypeScript or Haskell.

opensaas-sh/tools/dope.sh
```diff
@@ -39,6 +39,25 @@ ACTION=$3
...
+files_are_equal() {
+  local file1=$1
+  local file2=$2
+  if cmp -s "$file1" "$file2"; then
+    return 0
+  else
+    return 1
+  fi
+}
+
```
Martinsos, line 49: Why not just

```suggestion
file_is_binary() {
  file --mime "$1" | grep -q "charset=binary"
}

files_are_equal() {
  cmp -s "$1" "$2"
}
```

I don't think you loose anything by not having inputs named, it is quite clear from the name of the function and the code inside what those are since they are so super simple. And also mapping all non 0s to 1 is not something we really n eed, you might even say we are loosing information that way, so havig that if-then that maps to 0 or 1 is redundant.

Martinsos: Also, I believe you can put these at the bottom of the file, out of the way, since they are helper utility functions.

opensaas-sh/tools/dope.sh
Martinsos: Although I am looking at this specific comment now hah and ok this one is potentially maybe a bit reudndant, I mean it is quite clera we are generating  a .copy file from the lines below. But ok, I still don't mind, it is just that this is not the prime example, in my mind, of when it makes sense to put extra comments in bash. Because there is "Genreated ...copy" below. But I guess it helps a bit when it is on top like this. Yeah ok.

Martinsos (approved):
Looking good! I approved, but please check some of the comments, there are some minor improvements -> once you are happy with those, feel free to merge, I don't have to look agian.

opensaas-sh/tools/dope.sh
cprecioso: > you can put these at the bottom of the file, out of the way, since they are helper utility functions.

I don't think so, no? Shell needs the functions defined before usage, right? https://unix.stackexchange.com/questions/449498/call-function-declared-below

cprecioso: > Why not just

AFAIU we need to return the input code, not the output of the function. I didn't want to do the somewhat-weird `echo $?` incantation, preferred being specific.

I'll test out both things in case I was mistaken and it works nonetheless.

cprecioso: Bash does return the status code of the last run command inside a function, so the `if/return/else/return` was indeed excessive

opensaas-sh/tools/dope.sh
cprecioso: took them out in the end, with the code simplification it is right on the limit between yes or no so we can go the way of taking it out

opensaas-sh/tools/dope.sh
Martinsos: > > you can put these at the bottom of the file, out of the way, since they are helper utility functions.
> 
> I don't think so, no? Shell needs the functions defined before usage, right? https://unix.stackexchange.com/questions/449498/call-function-declared-below

Well, I remember that I defined utility functions at the bottom of installer.sh and then used them above. Ok oh, you are right, I just checked, so what I had there was this `main` function at top and then that was called at the very bottom, ok makes sense :). It is interpreted language with no second interpreter pass, so it can't know about func definitions if they aren not interpreted before the calls.

PR #526, opened 2025-10-02
Remove Satoshi font from template

### LINKED ISSUE #516: Reduce the `template` size

cprecioso:
- Closes #516
- On top of #525

Removes the Satoshi fonts from the template, as they were a leftover from pre-2.0, and maintains them in `opensaas-sh`, which does actually use Satoshi. We only keep the `.woff2` files as they were the only ones being used.


infomiho (approved):
I think this is a good change

PR #528, opened 2025-10-02
Add note about template requirements

cprecioso:
Based on https://github.com/wasp-lang/wasp/pull/3196#discussion_r2398305684

Apparently I left some stuff unformatted earlier so it also has those formatting things

### DIFF
--- a/.github/workflows/template-release.yml
+++ b/.github/workflows/template-release.yml
@@ -1,15 +1,15 @@
 ---
 name: Template Release
 
-'on':
+"on":
   push:
     tags:
-      - 'wasp-v*-template'
+      - "wasp-v*-template"
 
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/template-release.yml
```diff
@@ -30,6 +30,7 @@ jobs:
           ref: ${{ env.tag_name}}
 
       - name: Create template archive
+        # The Wasp CLI expects the template contents to be straight in the root of the archive.
```
infomiho:
```suggestion
        # The Wasp CLI expects the template contents to be in the root of the archive.
```

infomiho: Nitpick: I found the formulation awkward - maybe that's just me.

infomiho (approved):
Small suggestion

Issue #523, opened 2025-10-02
Improve documentation and investigate behavior of `dope.sh`


sodic:
There are several strange and undocumented behaviors with `dope.sh`. Some are desirable, others likely aren't.

1. We've designed `dope.sh` to ignore `package-lock.json` in the template directory. This was done on purpose because `package-lock.json` is only useful for us and we don't ship it as a part of the template.
2. It ignores untracked files. I am not sure if this in on purpose or an accidental side effect. In any case, we should document it in the README.
3. Problems emerge when you rename a file that's in the diff. Doing so creates conflicts that `dope.sh` doesn't know how to resolve. We should investigate this and fix it or document it.

PR #530, opened 2025-10-07
Use `dope.sh` to generate a testable app from production template

### LINKED ISSUE #524: Make it easier to test the Open Saas template in build mode
When releasing Wasp 0.18.0, I was tasked with testing Open Saas template in build mode.

I had to go through a series of steps to get it working:
1. I had to change the email from `Dummy` to something else, didn't know what at first.
2. I pulled the env for `opensaas-sh`, there are SendGrid credentials in there but I couldn’t get SendGrid to work.
   
    Image 
      Funnily enough, the UI said everything was OK:<br>

  Image

3. It was hard to figure out which provider we should set to use this credentials (SendGrid or SMTP). After setting the credentials, I couldn't (or didn't know how to) access SendGrid's dashboard to see what wasn't working. 
4. I tried to use SMTP credentials from `waspc/todoApp`. At first, they didn’t work for Open Saas despite working in `todoApp`, I had to remove the single quotes:
    
    ```
    # Had to change this
    SMTP_HOST='smtp.mailgun.org'
    SMTP_USERNAME='todoapp@mg.wasp.sh'
    SMTP_PASSWORD='<redacted>'
    SMTP_PORT=587
    
    # Into this
    SMTP_HOST=smtp.mailgun.org
    SMTP_USERNAME=todoapp@mg.wasp.sh
    SMTP_PASSWORD='<redacted>'
    SMTP_PORT=587
    ```
    
5. Login/registration still didn’t work after this, I had to change the email address from `me@example.com` to `todoapp@mg.wasp.sh` in two places inside `main.wasp`. It worked after that.
6. It’s hard to figure out which environment variables we should set (no dedicated `.env.` file we can pull)

infomiho:
## Description

Based on https://github.com/wasp-lang/open-saas/pull/531

Fixes #524 

This PR reuses the existing `dope.sh` to create a testable app from an app generated by `wasp new app -t saas`.

The changes are minimal:
- change `Dummy` to `SMTP` and use the correct email address
- setup Dotenv Vault

This allows for testing the app in production mode e.g. `npx @wasp.sh/wasp-app-runner build`.

This PR doesn't setup Stripe, analytics or similar. I think it's not really needed to smoke test the template. We have `opensaas.sh` with all the bell and whistles for that.

## Contributor Checklist

### DIFF
--- a/template-test/README.md
+++ b/template-test/README.md
@@ -20,7 +20,14 @@ cd app
 # Get the environment variables from Dotenv Vault.
 npm run env:pull
 
+# Run the dev database and create migrations.
+wasp db start
+wasp db migrate-dev --name "init"
+
+# You can stop the dev database after you created the migration.
+
 # Test that the app works in dev and prod modes.
+# Make sure the `@wasp.sh/wasp-app-runner` is updated to the latest version.
 npx @wasp.sh/wasp-app-runner dev
 npx @wasp.sh/wasp-app-runner build
 ```


cprecioso:
@infomiho haven't looked at anything here, but wanted to pop in; if we're going to be using `dope.sh` more extensively, might it make sense to migrate it to a more understandable (and less edge-casey) language? We can take advantage of the fact that node is already installed...

infomiho:
@cprecioso I think that the natural journey of all of our scripts -> inline -> bash -> Node.js

I wouldn't do it in this PR though, I'd create an issue if we agree that we want to do it.

cprecioso:
@infomiho  My vote is to even do that before this PR here. As you see fit.

Martinsos:
+1 for the idea of rewrite at some point but also big +1 for not doing it now, priorities again, dope.sh works fine and is tested through use, no need to shake that tree at the moment.

template-test/tools/diff.sh
sodic, line 1: We have a very similar thing in `opensaas-sh/tools`. Can we consolidate them and somehow?

template-test/app_diff/migrations/20251007102522_init/migration.sql.diff
sodic, line 1: I'm not sure about the migration stuff. Shouldn't `wasp db migrate-dev` be part of testing too?

template-test/app_diff/.gitignore.diff
sodic, line 1: Why does this file have diffs. I'd expect it to be the same (apart maybe from `!.env.valut`):
- Shouldn't `.env*` be ignored in the template too?
- Why is .flaskenv in there?
- Why is env.project in there?

template-test/app_diff/deletions
sodic, line 1: Do we need to delete these?

template-test/.gitignore
```diff
@@ -0,0 +1,2 @@
+app/
+base-app/
```
sodic: Why is this here? How would the `base-app` end up in this folder?

template-test/README.md
sodic, line 1: We should point the releaser to thsi from somwhere global (like the README.md or the release checklist).

template-test/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+# Test that the app works in dev and prod modes.
+npx @wasp.sh/wasp-app-runner dev
+npx @wasp.sh/wasp-app-runner build
+```
+
+### Workflow
+
+- Generate `app/` from template and diffs: `./tools/patch.sh`
+- Update diffs after modifying `app/`: `./tools/diff.sh`
```
sodic: I guess this is workflow for updating/migrating.

Since the dope.sh README is now in a different place, can we elaborate that here? You can put it under the title: "If you need to update/migrate..."

sodic (commented):
This is great! It solves the most annoying and the longest part of release testing (at least to me).

I do have some questions and suggestions, check the comments.

template-test/app_diff/.gitignore.diff
infomiho: We had this problem in different example apps - this is what Dotenv Vault automatically adds **each time** when you run `npm run env:pull`.

So if you are testing something and decide to modify the app e.g. you found some tweak that would help and run `./tools/diff.sh` you'll end up with the `.gitignore` changes again. You now you have to revert the `.gitignore` changes manually if you want to keep it out of the diff. 

So I decided to keep them in - they don't any harm and make the process less annoying.

template-test/app_diff/deletions
infomiho: I think I did by accident - I was probably cleaning up stuff I thought we didn't need or something like that. I'll remove this 👍

infomiho: Correction - it looks like a bug with the `dope.sh` script, I'll investigate!

template-test/tools/diff.sh
infomiho: They are actually quite different:
- `opensaas-sh/tools` script can use the `template` dir it has available on the disk
- `template-test/tools`script needs to run `wasp new` each time to get the template

infomiho: Image

template-test/.gitignore
infomiho: It's the side-effect of running `wasp new` to generate a new app based on the released Open Saas template. We decided to create the app in the `base-app` dir which we extract the `app` dir from. We clean it up - but just in case, I've gitignored it.

template-test/app_diff/migrations/20251007102522_init/migration.sql.diff
infomiho: Fair point! I'll remove the migrations from the diff since the tester needs to run `wasp db migrate-dev`.

template-test/app_diff/.gitignore.diff
sodic: Ok, fair enough, but can you put in a comment that explains this in the gitignore (basically what you said here).

template-test/app_diff/migrations/20251007102522_init/migration.sql.diff
infomiho: Good thing is that `wasp-app-runner` runs the `wasp db migrate-dev` command 😄 

Again the trouble is - every time you test the app and you tweak something, the new diff will include the migrations again. So you'll need to delete the diff files manually. What do we think about that?

template-test/app_diff/deletions
sodic: You might find this helpful: https://github.com/wasp-lang/open-saas/issues/523

template-test/.gitignore
sodic: Hm, I don't think we should ignore it then.

If it doesn't get cleaned up, we want to see problems at their true source?

template-test/app_diff/deletions
infomiho: False alarm, it looks the Dotenv Vault addition of `.env*` below the ` !.env.*.example` overrode the exception not to ignore those files - so it looked like the files were deleted when `dope.sh` ran. 

I've asked Claude to explain with code:

<details>
<summary>Explanation by Claude</summary>

## Root Cause: `.gitignore` Rule Ordering

In `template-test/app/.gitignore`, negation rules were placed **before** broader ignore patterns:

```bash
.env.*
!.env.*.example    # Un-ignores .env.*.example

.env*              # Re-ignores everything starting with .env (including .env.*.example)
!.env.vault        # Un-ignores .env.vault
```

**Last matching pattern wins in gitignore.**

## How dope.sh Detects Deletions

```bash
# dope.sh uses git ls-files to list files
git ls-files --cached --others --exclude-standard

# Files in BASE but not in DERIVED go into deletions
comm -23 <(echo "${BASE_FILES}") <(echo "${DERIVED_FILES}") > deletions
```

With broken gitignore:
```bash
cd template-test/app
git ls-files --cached --others --exclude-standard | grep "\.env"
# Output: .env.vault
# Missing: .env.client.example, .env.server.example
```

## The Fix

Reorder gitignore so negations come **after** broader patterns:

```bash
.env.*
.env*              # Ignores everything
!.env.vault        # Un-ignores specific files
!.env.*.example    # Un-ignores .env.*.example files
```

After fix:
```bash
git ls-files --cached --others --exclude-standard | grep "\.env"
# Output:
# .env.client.example
# .env.server.example
# .env.vault
```

Both BASE and DERIVED now have the files, so they don't appear in deletions.
</details>

template-test/README.md
infomiho: Linked to the README from the checklist https://www.notion.so/wasp-lang/0-X-Y-26818a74854c8094b1fdf42b3e9d25b8?source=copy_link#26818a74854c80439e0ae8bdd709b4d6

template-test/app_diff/.gitignore.diff
infomiho: Added a comment 👍

template-test/app_diff/deletions
infomiho: Now the file is empty - after I fixed the order in `.gitignore` 👍

template-test/app_diff/migrations/20251007102522_init/migration.sql.diff
infomiho: I've now removed `migrations` dir (template users start without them) so we can test `wasp db migrate-dev`. I've also removed the `package-lock.json` (template users start without it) so we can see test the template as our users would use it. 

Added extra lines in the `.gitignore` diff with explanations.

template-test/README.md
infomiho: I'm not sure where this went, now that I checked what I linked again. Could you please update the "in progress" checklist yourself to point to: https://github.com/wasp-lang/open-saas/blob/main/template-test/README.md

(Links to `main` branch)

sodic: Done!

template-test/tools/diff.sh
sodic: Ok, fair enough.

I thought the differences were maybe coming from us forgetting to do some stuff in one of the scripts (e.g., the `trap` thing). But if they're that different because they really are different, then all good.

RAW.

Issue #544, opened 2025-11-12
Implement a wrokflow to ensure `opensaas-sh` diffs are up to date


FranjoMindek:
Implement a workflow which would check if generating a diff makes changes to the working directory. If it does it fails the PR. 
This it to make sure our diffs are up to date with the template changes.

Read more here: https://github.com/wasp-lang/open-saas/pull/505#discussion_r2509900453

FranjoMindek:
Requires https://github.com/wasp-lang/open-saas/issues/548 to be implemented first.

PR #545, opened 2025-11-12
Update diffs to current template

FranjoMindek:
I run on `main` branch:

```sh
./opensaas-sh/tools/patch.sh
./opensaas-sh/tools/diff.sh
```

And committed the diffs.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -5,12 +5,10 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
 
 env:
   WASP_TELEMETRY_DISABLE: 1
-  WASP_VERSION: 0.18.0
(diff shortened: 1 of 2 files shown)


cprecioso:
@FranjoMindek there are no functional changes right? i think i ran into a case where `gpatch` on mac was creating different (but functionally identical) patch files than `patch` on linux

FranjoMindek:
Yes. My other PR had diffs unrelated to the PR so I wanted to make them up to date.

>  think i ran into a case where gpatch on mac was creating different (but functionally identical) patch files than patch on linux

Oh damn. 
I did plan to do this: https://github.com/wasp-lang/open-saas/issues/544
Hope this doesn't interfere.

Yep it does: https://github.com/wasp-lang/open-saas/actions/runs/19295088880/job/55174959604?pr=546

PR #546, opened 2025-11-12
Create a workflow to which stops uncommitted `opensaas-sh` diffs

FranjoMindek:
Solves: https://github.com/wasp-lang/open-saas/issues/544


FranjoMindek:
Will be handled here https://github.com/wasp-lang/open-saas/pull/557

Issue #543, opened 2025-11-12
Modernize Lemon Squeezy payment provider


FranjoMindek:
We want to revisit our implementation of Lemon Squeezy payment provider and implement all of the new practices we did when modernizing the Stripe payment provider.

Those include logging and webhook payload parsing logic.

More more about webhook payload parsing discussion here: https://github.com/wasp-lang/open-saas/pull/505#discussion_r2489729287

Issue #502, opened 2025-09-12
Reconsider `SubscriptionStatus` and it's handlers


FranjoMindek:
Now that we have access to more payment processors we should reconsider `SubScriptionStatus` implementation.
One good thing would be to better generalize our status names.

Currently we have:
```ts
export enum SubscriptionStatus {
  PastDue = 'past_due',
  CancelAtPeriodEnd = 'cancel_at_period_end',
  Active = 'active',
  Deleted = 'deleted',
}
```

One problem we have is that we have a state called `CancelAtPeriodEnd` but then that leads to `Deleted` rather than `Canceled` state.
This is also problematic if we want to support both `Canceled` and `Deleted` states in the future, where `Deleted` could be e.g. forced deletion of the subscription status like in Stripe sandbox mode.

FranjoMindek:
We should also reconsider how we handle `SubscriptionStatus` transitions.
We have a start of an idea for the current implementation here: https://github.com/wasp-lang/open-saas/pull/569#discussion_r2539408821

PR #569, opened 2025-11-14
Abstract Stripe payment logic into reusable functions

FranjoMindek:
Built on top of https://github.com/wasp-lang/open-saas/pull/570

Abstract logic that will be commonly used in all of payment providers out of Stripe specific logic.
Makes Stripe use that instead.

Note: Lemon Squeezy has some different approaches that will need to be fixed first before we can also use that logic in there. So for now I'm skipping it. That will be done as part of #543

### DIFF
--- a/template/app/src/payment/plans.ts
+++ b/template/app/src/payment/plans.ts
@@ -69,8 +69,8 @@ export function getSubscriptionPaymentPlanIds(): PaymentPlanId[] {
 }
 
 /**
- * Returns Open SaaS `PaymentPlanId` for some payment providers's plan ID.
- *
+ * Returns Open SaaS `PaymentPlanId` for some payment provider's plan ID.
+ * 
  * Different payment providers track plan ID in different ways.
  * e.g. Stripe price ID, Polar product ID...
  */


template/app/src/payment/stripe/webhook.ts
```diff
@@ -160,12 +164,12 @@ async function handleCustomerSubscriptionUpdated(
   }
 
   const customerId = getCustomerId(subscription.customer);
-  const paymentPlanId = getPaymentPlanIdByPriceId(
+  const paymentPlanId = getPaymentPlanIdByPaymentProcessorPlanId(
```
infomiho, line 167: Something feels off with the naming there:
`getPaymentPlanIdByPaymentProcessorPlanId` is called with `getSubscriptionPriceId` result as argument.

`getSubscriptionPriceId` should return some sort of subscription price ID

but `getPaymentPlanIdByPaymentProcessorPlanId` expects a payment processor **plan** ID?

template/app/src/payment/plans.ts
```diff
@@ -67,3 +68,15 @@ export function getSubscriptionPaymentPlanIds(): PaymentPlanId[] {
     (planId) => paymentPlans[planId].effect.kind === "subscription",
   );
 }
+
+export function getPaymentPlanIdByPaymentProcessorPlanId(
+  paymentProcessorUserId: NonNullable<User["paymentProcessorUserId"]>,
+): PaymentPlanId {
+  for (const [planId, plan] of Object.entries(paymentPlans)) {
+    if (plan.getPaymentProcessorPlanId() === paymentProcessorUserId) {
```
infomiho: This is again suspicious, you are comparing a payment plan ID with user ID?

template/app/src/payment/user.ts
```diff
@@ -39,54 +37,54 @@ export function updateUserPaymentProcessorUserId(
...
 
-interface UpdateUserOneTimePaymentDetailsArgs {
-  customerId: Stripe.Customer["id"];
-  datePaid: Date;
-  numOfCreditsPurchased: number;
+interface UpdateUserSubscriptionArgs {
+  paymentProcessorUserId: NonNullable<User["paymentProcessorUserId"]>;
+  subscriptionStatus: SubscriptionStatus;
+  paymentPlanId?: PaymentPlanId;
+  datePaid?: Date;
```
infomiho, line 44: As I understood the changes in this file, you just renamed the two functions and moved them around?

I'd also suggest going into analysis if we can avoid optional args here for different scenarios e.g. subscription deleted is the only case where we don't require the `paymentPlanId` and `datePaid`. Please analyze this and report back - I hope it's not too complex of a change (e.g. having 2 specific functions instead of one - `updateSubscription` and `markSubscriptionAsDelete` or smth like that)

template/app/src/payment/stripe/webhook.ts
FranjoMindek: Different payment providers recommend different ways to track "products".
"Payment process plan ID" is Open SaaS abstraction of that ID we need to track.

For Stripe you should be tracking `priceId` of some "product".
Open SaaS `paymentProcessorPlanId` == Stripe `priceId`, but 
Open Saas `paymentProcessorPlanId` == Polar `productId`

This is old code I wasn't part of when it was made.
We may want to rename Open SaaS abstractions at some point.

template/app/src/payment/plans.ts
FranjoMindek: This is a naming error. Sorry about that. I think i just read `paymentProcessor...Id`  when copy pasting from some other code and didn't realize it said `User` instead of `Plan`.

This is the right version:
```ts
export function getPaymentPlanIdByPaymentProcessorPlanId(
  paymentProcessorPlanId: string,
): PaymentPlanId {
  for (const [planId, plan] of Object.entries(paymentPlans)) {
    if (plan.getPaymentProcessorPlanId() === paymentProcessorPlanId) {
      return planId as PaymentPlanId;
    }
  }

  throw new Error(`Unknown payment processor plan ID: ${paymentProcessorPlanId}`);
}
```

template/app/src/payment/user.ts
FranjoMindek: I did this once already and deemed it too complicated.

We have 4 states of `SubscriptionStatus`:

| New `SubscriptionStatus`          | Fields to Update                                | Notes                                                                 |
|----------------------------------|--------------------------------------------------|-----------------------------------------------------------------------|
| `SubscriptionStatus.Active`      | `subscriptionPlan`, `subscriptionStatus`, `datePaid` |                    |
| `SubscriptionStatus.CancelAtPeriodEnd` | `subscriptionStatus`                          |                                     |
| `SubscriptionStatus.PastDue`     | `subscriptionStatus`, `subscriptionPlan` (conditional) | Plan changes depend on business rules (grace period, restricted access, remove plan). |
| `SubscriptionStatus.Deleted`     | `subscriptionPlan`, `subscriptionStatus`         | I think curent Open SaaS does not remove on deletion, which is a mistake in my opinion.                                    |

We can give some sensible default for `PastDue` but we would have to pick which one.
Because needs change with every status change, ideally it would be a handler per state, so N states = N functions.

I'm up for this, but we need to agree on changes, as they are breaking changes.
IMO I would implement it as outlined above + `grace period` solution on `PastDue`, but then I would have to implement logic for that. e.g. Wasp job
Easier one to do would be `remove plan`.

template/app/src/payment/stripe/webhook.ts
infomiho: Fair, I'd then introduce this explanation on top of this function. For any future readers that might get confused by the naming here: "different payment providers name stuff differently, we call it like this ..."

template/app/src/payment/user.ts
infomiho: I'd encode these ideas into future refactoring issue. This is a good start IMHO, if we can further clean up the logic, naming - that would be great. It's worthwhile doing since it's important payments code.

FranjoMindek: I've also recorded this conversation here https://github.com/wasp-lang/open-saas/issues/502#issuecomment-3558482722.

PR #594, opened 2026-01-09
Migrate to Wasp 0.21.0 (Tailwind 4)

infomiho:
## Description

Migrating Open SaaS to use Tailwind 4:
- did the migration steps
- migrated the old ShadCN components not to use `forwardRef`

Left to do:
- [x] Update the diffs, there are some merge conflicts.

## Contributor Checklist

### DIFF
--- a/opensaas-sh/app_diff/src/landing-page/components/Hero/Hero.tsx.diff
+++ b/opensaas-sh/app_diff/src/landing-page/components/Hero/Hero.tsx.diff
@@ -13,14 +13,14 @@
 +  const { data: user } = useAuth();
 +
 +  return (
-+    <div className="relative w-full [overflow-x:clip] pt-12">
++    <div className="relative w-full overflow-x-clip pt-12">
 +      <TopGradient />
 +      <BottomGradient />
 +      <div className="mx-auto flex max-w-7xl flex-col pt-20 lg:flex-row">
 +        <div className="py-24 sm:py-32">
 +          <div className="max-w-8xl px-6 lg:px-8">
 +            <div className="lg:mb-18 mx-auto max-w-3xl text-center md:text-left">
-+              <h1 className="text-foreground text-5xl font-extrabold md:text-6xl">
++              <h1 className="text-foreground text-5xl font-extrabold md:text-6xl leading-none">
 +                The <span className="italic">free</span> SaaS template with{" "}
 +                <span className="text-gradient-primary font-black">
 +                  superpowers
@@ -69,7 +69,7 @@
 +      aria-hidden="true"
 +    >
 +      <div
-+        className="aspect-[1020/880] w-[70rem] flex-none bg-gradient-to-tr from-amber-400 to-purple-300 opacity-10 sm:right-1/4 sm:translate-x-1/2 dark:hidden"
++        className="aspect-1020/880 w-[70rem] flex-none bg-linear-to-tr from-amber-400 to-purple-300 opacity-10 sm:right-1/4 sm:translate-x-1/2 dark:hidden"
 +        style={{
 +          clipPath:
 +            "polygon(80% 20%, 90% 55%, 50% 100%, 70% 30%, 20% 50%, 50% 0)",
@@ -86,7 +86,7 @@
 +      aria-hidden="true"
 +    >
 +      <div
-+        className="relative aspect-[1020/880] w-[90rem] bg-gradient-to-br from-amber-400 to-purple-300 opacity-10 sm:-left-3/4 sm:translate-x-1/4 dark:hidden"
++        className="relative aspect-1020/880 w-[90rem] bg-linear-to-br from-amber-400 to-purple-300 opacity-10 sm:-left-3/4 sm:translate-x-1/4 dark:hidden"
 +        style={{
 +          clipPath: "ellipse(80% 30% at 80% 50%)",
 +        }}
--- a/template/app/src/client/Main.css
+++ b/template/app/src/client/Main.css
@@ -147,14 +147,6 @@
   }
 }
 
-/*
-  The default border color has changed to `currentcolor` in Tailwind CSS v4,
-  so we've added these compatibility styles to make sure everything still
-  looks the same as it did with Tailwind CSS v3.
-
-  If we ever want to remove these styles, we need to add an explicit border
-  color utility to any element that depends on these defaults.
-*/
 @layer base {
   *,
   ::after,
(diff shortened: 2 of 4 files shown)


template/app/src/landing-page/components/Testimonials.tsx
```diff
@@ -31,7 +31,7 @@ export default function Testimonials({
     <div className="mx-auto mt-32 max-w-7xl sm:mt-56 sm:px-6 lg:px-8">
       <SectionTitle title="What Our Users Say" />
 
-      <div className="relative z-10 w-full columns-1 gap-2 px-4 md:columns-2 md:gap-6 md:px-0 lg:columns-3">
+      <div className="relative z-10 w-full columns-1 gap-2 px-4 md:columns-2 md:gap-6 md:px-0 lg:columns-[3]">
```
infomiho: For some wild reason, `columns-3` returns `column: 3` but `lg:columns-3` returns `column: 0.75em`. I had to use `lg:columns-[3]` to make it behave as expected. 

I think I found a bug in Tailwind's upgrade tool, reported it here: https://github.com/tailwindlabs/tailwindcss/issues/19555

infomiho: I've removed our custom `--container-*` vars since all the `max-w-*` classes work find without them? I believe it's due to TW4 being able to handle any suffix and just uses it to multiply the number in e.g. `max-w-2.5` in `var(--spacing) * 2.5`.

opensaas-sh/app_diff/.env.vault.diff
cprecioso, line 1: I guess this is just a format update by dotenv, this resolves to the same secrets right? Just double-checking.

template/app/src/client/components/ui/accordion.tsx
```diff
@@ -1,55 +1,64 @@
...
 import { ChevronDown } from "lucide-react";
-import * as React from "react";
 
 import { cn } from "../../utils";
 
-const Accordion = AccordionPrimitive.Root;
+function Accordion({
+  ...props
+}: React.ComponentProps<typeof AccordionPrimitive.Root>) {
+  return <AccordionPrimitive.Root data-slot="accordion" {...props} />;
```
cprecioso, line 10: What are these `data-slot` things? It doesn't seem like Radix needs them.

template/app/src/client/Main.css
```diff
@@ -1,105 +1,268 @@
...
+  looks the same as it did with Tailwind CSS v3.
 
-.tableCheckbox:checked ~ div span {
-  @apply opacity-100;
-}
-.tableCheckbox:checked ~ div {
-  @apply bg-primary border-primary;
+  If we ever want to remove these styles, we need to add an explicit border
+  color utility to any element that depends on these defaults.
+*/
```
cprecioso: I'd remove this comment, OpenSaaS users don't know / need to know that we updated from TW 3 to 4.
Then I'd evaluate if this snippet is needed or not. (e.g. I removed it in all of our example apps).

template/app/src/client/Main.css
```diff
@@ -1,105 +1,268 @@
...
+  .apexcharts-yaxistooltip {
+    @apply dark:bg-muted! dark:border-muted! dark:text-muted-foreground!;
+  }
+  .apexcharts-xaxistooltip-bottom:after {
+    @apply dark:border-b-muted!;
+  }
+  .apexcharts-xaxistooltip-bottom:before {
+    @apply dark:border-b-muted!;
+  }
 }
```
cprecioso: I'm not sure of "the tailwind way", but these are not utilities, no? They are not building blocks for larger styles; but specific names we have to set to customize apex charts. Then I don't think they should be in the `utilities` layer, but in the default nameless layer, as every other application style.

template/app/src/client/components/ui/accordion.tsx
cprecioso: Checking my understanding: all the changes in `components/ui` are the components from Shadcn being regenerated, right? The `data-slot` thing is a Shadcn internal then?

opensaas-sh/app_diff/.env.vault.diff
infomiho: Someone updated the `.env.server` contents and pushed it to the Dotenv Vault, but didn't commit this change. So when I pulled the latest version, I got this diff.

template/app/src/client/components/ui/accordion.tsx
infomiho: Yep, the latest version of ShadCN uses those slots, so we have them in our code now. It's maybe redundant but it makes our code look more similar to what is out there:

image

template/app/src/client/Main.css
infomiho: This was done by the migration tool, I didn't even think about it. You have a good point 👍

template/app/src/client/Main.css
infomiho: This comment was also added by the migration tool lol https://github.com/tailwindlabs/tailwindcss/blob/main/packages/%40tailwindcss-upgrade/src/codemods/css/migrate-preflight.ts#L16

I didn't even catch it was written in such a human way. I agree with you, I'll remove it 👍

opensaas-sh/app_diff/src/landing-page/components/Hero/Hero.tsx.diff
```diff
@@ -13,14 +13,14 @@
...
-+    <div className="relative w-full [overflow-x:clip] pt-12">
++    <div className="relative w-full overflow-x-clip pt-12">
 +      <TopGradient />
 +      <BottomGradient />
 +      <div className="mx-auto flex max-w-7xl flex-col pt-20 lg:flex-row">
 +        <div className="py-24 sm:py-32">
 +          <div className="max-w-8xl px-6 lg:px-8">
 +            <div className="lg:mb-18 mx-auto max-w-3xl text-center md:text-left">
-+              <h1 className="text-foreground text-5xl font-extrabold md:text-6xl">
++              <h1 className="text-foreground text-5xl font-extrabold md:text-6xl leading-none">
```
infomiho, line 23: I had to add this because the `line-height` value of `1` from `text-6xl` was overridden in TW4 by the `leading-tight` value of `1.25` (which we had set for all headings). 

| Image | Image |
| --- | --- | 
| Before | After |

infomiho:
@cprecioso ready for another look

template/app/src/client/Main.css
cprecioso: > This comment was also added by the migration tool lol

Yep yep i know, I had it in all of our example apps migrations.

Issue #601, opened 2026-01-15
add docs info for CC plugin


vincanger:
we now have a wasp cc plugin https://github.com/wasp-lang/claude-plugins

wang2-lat:
Nice! You can add a section in the main README linking to the Claude plugin repo, or create a `docs/integrations.md` file to list it alongside other IDE/editor plugins. Would help with discoverability.

vincanger:
resolved in #606

PR #606, opened 2026-01-15
update AI info (skills, AGENTS.md, etc.)

### LINKED ISSUE #584: Improve template directory with Claude Code integration documentation
Part of #579

## Feature Request

Enhance the `template/` directory to include documentation and resources that make working with Claude Code easier.

### Suggested improvements:
- Add Claude Code-specific documentation for the project
- Include setup guides for using Claude Code with this template
- Add relevant configuration files or examples
- Document common workflows when using Claude Code with this codebase

This would help developers using Claude Code to work more efficiently with the template.

vincanger:
## Summary
- Add skills and mcp settings based on agent standards (Agent Skills & AGENTS.md -> CLAUDE.md)
- Decided on copying the plugin skills to `.claude/skills` and using AGENTS.md instead of installing plugin (this means we would have to copy the plugin folders over to open saas whenever we update the CC plugin until we come up with a better way)
- Update vibe coding guide documentation
- Adjust LP messaging slightly

Fixes #584 and #601

### DIFF
--- a/opensaas-sh/blog/src/content/docs/guides/vibe-coding.mdx
+++ b/opensaas-sh/blog/src/content/docs/guides/vibe-coding.mdx
@@ -29,7 +29,7 @@ Rules, found within the `AGENTS.md` (and `CLAUDE.md`), is another [open standard
 :::tip[These are just a starting point!]
 Use these skills and rules to get your project up and running.
 
-Run the `add-wasp-skills` skill to add [Wasp framework-specific skills](https://github.com/wasp-lang/claude-plugins/tree/main/plugins/wasp) to your project for even more power.
+Run the `add-wasp-skills` skill to add [Wasp framework-specific skills](https://github.com/wasp-lang/wasp-agent-plugins/tree/main/plugins/wasp) to your project for even more power.
 :::
 
 ## LLM-Friendly Documentation
--- a/opensaas-sh/blog/src/content/docs/blog/2026-03-16-best-way-to-vibe-code-saas-2026.mdx
+++ b/opensaas-sh/blog/src/content/docs/blog/2026-03-16-best-way-to-vibe-code-saas-2026.mdx
@@ -202,18 +202,18 @@ This scaffolds a new Open SaaS project with auth, payments, email, file uploads,
 
 ### Step 2: Set Up Claude Code with the Wasp Plugin
 
-After [installing Claude Code](https://code.claude.com/docs/en/quickstart), add [the Wasp plugin](https://github.com/wasp-lang/claude-plugins/tree/main/plugins/wasp) so that it has deep knowledge of the framework, as well as skills and instructions on how to properly use full-stack debugging visibility, llms.txt docs, and more:
+After [installing Claude Code](https://code.claude.com/docs/en/quickstart), add [the Wasp plugin](https://github.com/wasp-lang/wasp-agent-plugins/tree/main/plugins/wasp) so that it has deep knowledge of the framework, as well as skills and instructions on how to properly use full-stack debugging visibility, llms.txt docs, and more:
 
 ```bash
-claude plugin marketplace add wasp-lang/claude-plugins
-claude plugin install wasp@wasp-plugins --scope project
+claude plugin marketplace add wasp-lang/wasp-agent-plugins
+claude plugin install wasp@wasp-agent-plugins --scope project
 ```
 
 :::note[Other coding agents]
 If you're using a coding agent tool other than Claude Code, like Codex, OpenCode, or Cursor, you can get the same set of agent skills by running:
 
 ```bash
-npx skills add wasp-lang/claude-plugins
+npx skills add wasp-lang/wasp-agent-plugins
 ```
 :::
 
--- a/template/app/.agents/skills/add-wasp-skills/SKILL.md
+++ b/template/app/.agents/skills/add-wasp-skills/SKILL.md
@@ -12,8 +12,8 @@ Let the user know that Wasp agent skills (also called plugins) are available for
 
 ## Step 2: Fetch Installation Instructions
 
-Fetch the installation instructions from: https://raw.githubusercontent.com/wasp-lang/claude-plugins/refs/heads/main/README.md
+Fetch the installation instructions from: https://raw.githubusercontent.com/wasp-lang/wasp-agent-plugins/refs/heads/main/plugins/wasp/README.md
 
 ## Step 3: Guide Installation
 
-Walk the user through the installation instructions from the fetched README. Follow them exactly as written.
+Walk the user through the installation instructions from the fetched README.


opensaas-sh/blog/src/content/docs/guides/vibe-coding.mdx
Martinsos: We are mixing ai and vibe coding here. Vibe means you are not checking the code at all. We should udpate the title and make it about AI coding, not vibe coding, since this is useufl for all peopel using AI (which is eveyrbody hah), not just "vibe" coders.

opensaas-sh/blog/src/content/docs/guides/vibe-coding.mdx
```diff
@@ -15,34 +13,48 @@ If you're looking to use AI to help build (or "vibe code") your SaaS app, this g
...
-Luckily, we did the work for you and put together a bunch of resources to help you use Wasp & Open SaaS with AI as effectively as possible.
+## Agent Skills and Rules
 
-### AI Resources in the Template
+We've packaged Wasp agent skills and rules into the template to be used by your favorite AI-assisted coding tools (Claude Code, Cursor, VS Code/Copilot, Codex, Gemini CLI, OpenCode, etc.).
 
-The template comes with:
-- A full set of rules files, `app/.cursor/rules`, to be used with Cursor or adapted to your coding tool of choice (Windsurf, Claude Code, etc.).
-- A set of example prompts, `app/.cursor/example-prompts.md`, to help you get started.
+Skills are found in the `app/.claude` directory of the Open SaaS template, but because [Agent Skills](https://agentskills.io/) are a recognized standard, your AI-assisted coding tool will either load them automatically from the `.claude` directory, or can be used as-is after they've been properly referenced.
```
Martinsos: what does it mean they can be used as-is after they've been properly referenced?

opensaas-sh/blog/src/content/docs/guides/vibe-coding.mdx
```diff
@@ -15,34 +13,48 @@ If you're looking to use AI to help build (or "vibe code") your SaaS app, this g
...
 
-<Image src={llmsTextChat} alt="add llms.txt to settings" />
-
-### More AI-assisted Coding Learning Resources
+- [Open SaaS Docs - LLMs.txt](https://docs.opensaas.sh/llms.txt) - Links to the raw text docs.
+- [Open SaaS Docs - LLMs-full.txt](https://docs.opensaas.sh/llms-full.txt) - Complete docs as one text file.
+- [Wasp Docs - LLMs.txt](https://wasp.sh/llms.txt) - Links to the raw text docs by Wasp version.
+- [Wasp Docs - LLMs-full.txt](https://wasp.sh/llms-full.txt) - Complete docs as one text file.
+
+## MCP Server Installation
```
Martinsos: Maybe explain what is this about, I feel it is missing context.

Martinsos:
@vincanger Cool effort here!

I tihnk we should do something like this, the main thing though is that it feels incosistent.

So here, we offer an AGENTS.md (rules) and .claude/skills (skills), so that anybody using OpenSaas, has an easier time using AI with it. Ok, motivation is clear.

Those however are not OpenSaas-specific rules or skills, but Wasp-specific ones, and they are the same as in our CC Wasp plugin. This is the first problem -> if they are not OpenSaaS speciifc, why are we offering them only in OpenSaas, why not in other templates (e.g. those taht come with Wasp)?
It also means they are out of Wasp's "lifecycle", so I can imagine it will be easy to froget to update them as we update Wasp.

Second, we are duplicating them from CC plugin, as you said your self is a problem. That will be tricky to maintain.

My thinking is, if we want to do this, let's do it right, so we can maintain it properly.
First of all, if these are not OpenSaaS specific, I wouldn't have them in OpenSaas.
OK, let's say we want to make it easy for people to get these in any Wasp project. Solution I can think of is that they run a command and bam, these are added to their Wasp project. Maybe `wasp ai:setup`? `wasp ai:init`? Something. So it could be part of Wasp CLI -> wasp produces useful AI rules and stuff for you. Maybe also informs you about CC plugin -> asks if you want CC plugin, or just copy-pasted stuff.
If not part of `wasp` CLI, it could be a separate thing, maybe a small npm package that one can run with `npx wasp-ai-setup`. Then however it can't be versionsed together with the CLI, but that might be just fine.
There is still a problem that we duplicate knowledge between CC Wasp plugin, and these standalone rules and skills. But maybe we don't have to -> can't we just copy them from CC Wasp plugin?

opensaas-sh/blog/src/content/docs/guides/vibe-coding.mdx
vincanger: I changed this up so this is no longer relevant.

opensaas-sh/blog/src/content/docs/guides/vibe-coding.mdx
vincanger: This is also no longer relevant. I changed this up by packing it into a skill so that it gets installed by the agent --> https://github.com/wasp-lang/claude-plugins/pull/23

vincanger:
Ok @Martinsos so here's the approach I went for so that we don't duplicate knowledge across CC plugin and repos:

- i made our CC plugin's skill descriptions agent agnostic in [this PR](https://github.com/wasp-lang/claude-plugins/pull/23) (e.g. replaced "claude" with "agent")
- added instructions to the CC plugin readme on how to use `npx skills add` to install skills for other agents (codex, open code, cursor)
 
- in this PR, I removed all skills that were duplicated from the CC plugin and just added a few agent agnostic Open SaaS starter skills such as `/getting-started` that pull the corresponding open saas doc guides and help them get wasp installed and the app running
- there's also an `/add-wasp-skills` for optionally adding the wasp plugin/skills if they want.

so now the only skills included in the open saas repo are unique to it, and they all defer to docs URLs and READMEs so there is very little to no maintenance required to maintain them.

opensaas-sh/app_diff/.claude/settings.json.diff
Martinsos, line 1: Just to check, this file still makes sense as it is?

opensaas-sh/blog/astro.config.mjs
```diff
@@ -76,7 +76,7 @@ export default defineConfig({
             { label: 'File Uploading', link: "/guides/file-uploading/" },
             { label: 'Tests', link: "/guides/tests/" },
             { label: 'How (Not) to Update Your Open SaaS App', link: "/guides/updating-opensaas/" },
-            { label: 'Vibe Coding with Open SaaS', link: "/guides/vibe-coding/" },
+            { label: 'Using AI Coding Tools', link: "/guides/vibe-coding/" },
```
Martinsos, line 93: Link is still vibe-coding though, maybe we want to change it to ai-coding?

template/app/.claude/skills/add-wasp-skills
Martinsos, line 1: What am I looking at here, this is a textual file with path to file two level above -> so this is claude specific skill, just pointing to a general skill. Does claude need this, it won't pick up stuff from .agents on its own? How is that this just being a relative path works? Is this even a textual file hm I guess it is, has no extension but maybe that is all right? Same goes for other two files.

template/app/AGENTS.md
```diff
@@ -0,0 +1,15 @@
+
+# Open SaaS 
+
+This is SaaS boilerplate starter kit built on top of Wasp, a batteries-included framework for building full-stack web apps with React, Node.js, and Prisma.
+
+## Documentation
+
+### Open SaaS
+Always fetch and verify your knowledge against the Open SaaS documentation before taking on tasks, answering questions, or doing any development work in this project:
+1. Fetch the Open SaaS documentation map from the [LLMs.txt index](https://docs.opensaas.sh/llms.txt). The map contains raw markdown file GitHub URLs of all documentation sections.
```
Martinsos, line 10: How does it know to fetch the correct version? What if project is older and OpenSaas changed quite a bit, I guess that could become quite confusing for the users? I am not sure what to do about it though. Would be cool if we could point it to a specific commit. Although llms.txt is not in the github is it, it is only deployed I think. And we don't deploy old versions for open saas. Hmmmmm tricky! Can we at least somehow warn it that this might happen so it can maybe handle taht somehow? Do you have any ideas?

opensaas-sh/app_diff/.claude/settings.json.diff
vincanger: Yeah. It's got good permissions for working in opensaas-sh, so I'd leave it.

opensaas-sh/blog/astro.config.mjs
vincanger: I'd leave it. So if people search `vibe coding saas` it might get a hit. In the last 28 days we've got some decent number of direct hits there (108 out of 259)

Image

template/app/.claude/skills/add-wasp-skills
vincanger: Each entry in `.claude/skills/` is a symlink pointing to the corresponding skill directory in .agents/skills/.                                                                             
                                                                                                                                       
So the actual skill content (the SKILL.md) lives in `.agents/`, which is the tool-agnostic location that works with Cursor, Windsurf, etc. The `.claude/skills/` symlinks just let Claude Code discover those same skills without duplicating them.

template/app/AGENTS.md
vincanger: They probably wouldn't get this file for older versions since it only comes with newer versions anyway.

We don't version Open SaaS docs, and to be honest, most of the valuable Open SaaS docs are about project structure (which LLMs can figure out on their own) and setting up integrations (which don't need to be versioned). The real important information an LLM would need would be Wasp docs then. So I don't really think there is a need to warn them either.

PR #608, opened 2026-01-16
Fix dark mode after Tailwind 4 upgrade

### LINKED ISSUE #607: Tailwind 4 upgrade broke dark mode
The upgrade to Tailwind 4 (#594) broke dark mode on the template and the OpenSaaS website.

Image
(This should be dark)

cprecioso:
- Fixes #607

## Summary
The main problem is that `@theme` by default tries to generate code into the classes directly, so indirections through CSS variables don't work. To fix it, you have to add the `inline` option so that theme is outputted itself and can follow CSS variables: https://tailwindcss.com/docs/theme#referencing-other-variables

Changes:
- Uses the dark mode selector as advised by Tailwind [(docs)](https://tailwindcss.com/docs/dark-mode#toggling-dark-mode-manually)
- Changes `@theme` to `@theme inline` for proper Tailwind 4 compatibility
- Normalizes quote style in CSS directives

### DIFF
--- a/template/app/src/client/Main.css
+++ b/template/app/src/client/Main.css
@@ -1,12 +1,12 @@
 @import "tailwindcss" source("..");
 
-@plugin '@tailwindcss/forms';
-@plugin '@tailwindcss/typography';
-@plugin 'tailwindcss-animate';
+@plugin "@tailwindcss/forms";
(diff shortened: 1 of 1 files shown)


infomiho (approved):
LGTM

PR #650, opened 2026-03-19
fix(payments): store actual subscription end date from payment providers

### LINKED ISSUE #407: Do a deep review of the payments logic
We should review our payments module and double check there's no funny business (specifically payment/subscription date).

Hassad674:
Fixes #407

  ## Audit findings

  Reviewed the entire payments module across all three providers. Found these date handling issues:

  1. **No `subscriptionEndDate` field** — the app had no way to store when the billing period actually ends. The `AccountPage` used `datePaid + 1 month` via `setMonth()` which has a JS month-overflow bug (Jan 31
   + 1 month = March 3, not Feb 28).

  2. **LemonSqueezy uses `new Date()` instead of payload timestamps** — Stripe and Polar correctly use timestamps from the webhook payload, but LemonSqueezy was using server time. If there's webhook delivery
  delay, the stored date would be wrong.

  3. **LemonSqueezy uses string literal instead of enum** — `"cancel_at_period_end" as SubscriptionStatus` instead of `SubscriptionStatus.CancelAtPeriodEnd`.

  4. **Stripe doesn't capture `current_period_end`** from `customer.subscription.updated` events.

  5. **Polar doesn't capture `currentPeriodEnd`** from `subscription.updated` events.

  6. **`subscriptionEndDate` not cleared on deletion** — stale data remains after subscription ends.

  ## Additional findings (not fixed, out of scope)

  - No webhook idempotency — retried webhooks could double-credit users (architectural change, deserves its own issue)
  - `subscriptionPlan` not cleared on deletion (cosmetic, no functional impact)
  - Analytics excludes `cancel_at_period_end` users from paid count (intentional per code comment)

  ## Changes

  - Added `subscriptionEndDate DateTime?` to User schema
  - Stripe: extract `current_period_end` from invoice and subscription webhooks
  - LemonSqueezy: use `created_at`/`updated_at` from payload, added fields to Zod schema
  - Polar: extract `currentPeriodEnd` from subscription webhooks
  - AccountPage: use real end date with fallback for existing users
  - Clean up `subscriptionEndDate` on subscription deletion
  - Updated seed data

  ## Tests

  - 12 unit tests (vitest) covering date extraction logic for Stripe, LemonSqueezy, and billing period display
  - 2 e2e tests (playwright) verifying account page displays subscription info correctly
  - Manual e2e test with real Stripe checkout confirmed `subscriptionEndDate` is stored and displayed

  ## Contributor Checklist

  - [x] **Update e2e tests**: Added `accountPageTests.spec.ts`
  - [x] **Update demo app**: Updated seed data with `subscriptionEndDate`

### DIFF
--- a/template/app/package.json
+++ b/template/app/package.json
@@ -53,6 +53,7 @@
     "@types/react": "^19.2.7",
     "prisma": "5.19.1",
     "typescript": "5.8.2",
-    "vite": "^7.0.6"
+    "vite": "^7.0.6",
+    "vitest": "^4.1.0"
   }
 }
--- a/template/app/schema.prisma
+++ b/template/app/schema.prisma
@@ -20,6 +20,7 @@ model User {
   subscriptionStatus        String?         // 'active', 'cancel_at_period_end', 'past_due', 'deleted'
   subscriptionPlan          String?         // 'hobby', 'pro'
   datePaid                  DateTime?
+  subscriptionEndDate       DateTime?
   credits                   Int             @default(3)
 
   gptResponses              GptResponse[]
--- a/template/app/src/payment/__tests__/billing-period-display.test.ts
+++ b/template/app/src/payment/__tests__/billing-period-display.test.ts
@@ -1,0 +1,56 @@
+import { describe, it, expect } from "vitest";
+
+// Mirrors the formatBillingPeriodEnd logic from AccountPage.tsx
+function formatBillingPeriodEnd(
+  datePaid: Date,
+  subscriptionEndDate: Date | null
+): string {
+  if (subscriptionEndDate) {
+    return subscriptionEndDate.toLocaleDateString();
+  }
+  const estimated = new Date(datePaid);
+  estimated.setMonth(estimated.getMonth() + 1);
+  return estimated.toLocaleDateString();
+}
+
+describe("formatBillingPeriodEnd", () => {
+  it("uses subscriptionEndDate when available", () => {
+    const datePaid = new Date("2025-01-15");
+    const endDate = new Date("2025-02-15");
+
+    const result = formatBillingPeriodEnd(datePaid, endDate);
+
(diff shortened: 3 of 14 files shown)


FranjoMindek:
Hey,

the changes are not too bad, in general they make sense.
The `oneMonthFromNow.setMonth(oneMonthFromNow.getMonth() + 1);` bug we have is visual only. But it exists.

As you've mentioned, truly fixing all of the problems would require great architectural changes.

I would be happy to accept the PR if keep in mind:

1. This is a template, so we don't really keep backwards compatibility. With your changes `formatBillingPeriodEnd` is unnecessary as we can just look directly into the field `subscriptionEndDate`. No fallback is necessary.
2. Cleaned up Stripe logic for getting `subscriptionEndDate`. Make it more similar to existing code.
3. We have existing playwrgiht test project. So e2e tests are unnecessary here.
4. You've dropped the unit tests in their current form. While they add some regressions they are not really useful when taking above points into consideration.

JiwaniZakir (commented):
The `billing-period-display.test.ts` file duplicates the `formatBillingPeriodEnd` logic inline rather than importing it from `AccountPage.tsx`. This means the tests validate a copy of the function, not the actual implementation — if the real function is updated, these tests will continue passing without catching the regression. The function should be extracted to a shared utility and imported in both the component and the test.

Additionally, `stripe-webhook.test.ts` imports `vi` from vitest but never uses it — worth cleaning up to avoid confusion about whether mocking was intended there. More broadly, the webhook test files (`lemonsqueezy-webhook.test.ts`, `stripe-webhook.test.ts`) test raw date parsing primitives in isolation rather than the actual handler functions, so they don't cover the code paths that changed (e.g., the `handleOrderCreated` fix in `webhook.ts` using `data.attributes.created_at` instead of `new Date()`). A test that exercises the actual handler with a mocked DB call would give much stronger confidence that the fix behaves correctly end-to-end.

PR #664, opened 2026-04-20
fix: render backtick-enclosed text as inline code in roadmap titles

### LINKED ISSUE #583: Backticks on Roadmap on landing page are messed up
Image

They appear on top of letters / misaligned for some reason.

octo-patch:
Fixes #583

## Problem

GitHub project titles can contain backtick characters (e.g., `` `Demo App` ``). When rendered as plain text in the Roadmap section of the landing page, the backticks appeared as misaligned diacritics (the character "Ì") on top of adjacent letters instead of rendering as inline code.

## Solution

Added a `renderTitle` helper function in `RoadmapEpicCard.tsx` that:
1. Splits the title on backtick pairs using a regex
2. Renders backtick-enclosed substrings as `<code>` elements with a subtle gray background styling consistent with the existing design
3. Returns plain strings unchanged (when no backticks are present) for minimal overhead

The result is that a title like `` Improve GPT Wrapper `Demo App` `` now shows "Demo App" in an inline code style, matching the intended Markdown semantics of GitHub project titles.

## Testing

The fix is contained to the `renderTitle` helper in `RoadmapEpicCard.tsx`. The logic can be verified by checking that:
- Titles with no backticks are returned as plain strings (no change in rendering)
- Titles with backtick pairs have the enclosed content wrapped in styled `<code>` elements
- Partial/unmatched backticks (odd number) are left as-is by the regex split

The change is in the `app_diff` file which gets applied to produce the `opensaas-sh` app.

### DIFF
--- a/opensaas-sh/app_diff/src/landing-page/components/RoadmapEpicCard.tsx.diff
+++ b/opensaas-sh/app_diff/src/landing-page/components/RoadmapEpicCard.tsx.diff
@@ -1,6 +1,6 @@
 --- template/app/src/landing-page/components/RoadmapEpicCard.tsx
 +++ opensaas-sh/app/src/landing-page/components/RoadmapEpicCard.tsx
-@@ -0,0 +1,92 @@
+@@ -0,0 +1,109 @@
 +import { Plus } from "lucide-react";
 +import { cn } from "../../client/utils";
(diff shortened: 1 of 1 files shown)


sodic:
Please attach a screenshot.

PR #670, opened 2026-05-04
Make payment URLs a bit more typesafe

### LINKED ISSUE #573: Reduce duplication on internal links

FranjoMindek:
Fixes #573 

Make checkout result and customer portal URLs a bit more typesafe.
Ideally we would support declaring search query types on routes.

Kinda like tanstack: https://tanstack.com/router/latest/docs/guide/search-params#validating-and-typing-search-params

### DIFF
--- a/opensaas-sh/app_diff/src/payment/stripe/paymentProcessor.ts.diff
+++ b/opensaas-sh/app_diff/src/payment/stripe/paymentProcessor.ts.diff
@@ -42,6 +42,6 @@
        await stripeClient.billingPortal.sessions.create({
 -        customer: paymentProcessorUserId,
 +        customer: stripeId,
-         return_url: `${config.frontendUrl}/account`,
+         return_url: CUSTOMER_PORTAL_RETURN_URL,
        });
  
--- a/template/app/src/payment/lemonSqueezy/checkoutUtils.ts
+++ b/template/app/src/payment/lemonSqueezy/checkoutUtils.ts
@@ -1,4 +1,5 @@
 import { createCheckout } from "@lemonsqueezy/lemonsqueezy.js";
+import { CHECKOUT_SUCCESS_URL } from "../paths";
 
 interface LemonSqueezyCheckoutSessionParams {
   storeId: string;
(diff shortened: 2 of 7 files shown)


template/app/src/payment/lemonSqueezy/checkoutUtils.ts
```diff
@@ -20,6 +21,9 @@ export async function createLemonSqueezyCheckoutSession({
         user_id: userId, // You app's unique user ID is sent on checkout, and it's returned in the webhook so we can easily identify the user.
       },
     },
+    productOptions: {
```
FranjoMindek, line 24: We didn't have this before 🤔 .

I should really refactor lemonsqueezy when I get time.

opensaas-sh/app_diff/src/payment/stripe/paymentProcessor.ts.diff
```diff
@@ -42,6 +42,6 @@
        await stripeClient.billingPortal.sessions.create({
 -        customer: paymentProcessorUserId,
 +        customer: stripeId,
-         return_url: `${config.frontendUrl}/account`,
+         return_url: CUSTOMER_PORTAL_RETURN_URL,
```
infomiho, line 45: If we can maybe come up with a URL which has a meaning in the Open SaaS universe e.g. ACCOUNT_PAGE and not in the Stripe universe e.g. CUSTOMER_PORTAL_RETURN_URL

infomiho (approved):
Just one naming suggestion

opensaas-sh/app_diff/src/payment/stripe/paymentProcessor.ts.diff
FranjoMindek: I want it to be in Stripe universe, mostly since this is located in `payments`, not the client.

Stripe only knows customer portal return url, and the file hides the connection to the client.

infomiho: But it's a constant referring to a concept in Open SaaS, the `return_url` is the border between two domains. So if you write code like: `return_url: ACCOUNT_PAGE_URL` you are saying "the return URL is the account page". Not a big deal 👍

FranjoMindek: I get it, but I'd rather payments not know concrete frontend architecture.
It knows `CUSTOMER_PORTAL_RETURN_URL`, be it account page, home page, settings page, it does not matter to payment processor.

Issue #662, opened 2026-04-14
Decide on prerendering strategy for opensaas


cprecioso:
We merged prerendering in Wasp `main`. We can now use it in OpenSaas. Let's decide in which pages!

Martinsos:
Consider recommending also using prerendering for blog posts and static pages so they odn't have ot use astro project on the side. That said, without route params supported in prerendering, they might still want to use astro for more complex stuff, let's see.

cprecioso:
Done in #677

PR #677, opened 2026-05-18
add prerendering to template and opensaas-sh

vincanger:
## Description

Fixes issue #662 by adding prerendering to the LP and /pricing for the template, but only to the LP for opensaas-sh
image

## Contributor Checklist


opensaas-sh/app_diff/main.wasp.diff
```diff
@@ -117,7 +121,16 @@
  //#region Auth Pages
  route LoginRoute { path: "/login", to: LoginPage }
  page LoginPage {
-@@ -211,9 +213,9 @@
+@@ -191,7 +191,7 @@
+ //#endregion
+ 
+ //#region Payment
+-route PricingPageRoute { path: "/pricing", to: PricingPage, prerender: true }
++route PricingPageRoute { path: "/pricing", to: PricingPage }
```
cprecioso, line 129: Was this intentional? I think we can keep the pricing page prerendered too

opensaas-sh/blog/astro.config.mjs
```diff
@@ -85,7 +85,7 @@ export default defineConfig({
               ],
             },
             { label: "Deploying", link: "/guides/deploying/" },
-            { label: "SEO", link: "/guides/seo/" },
+            { label: "SEO & Performance", link: "/guides/seo-performance/" },
```
cprecioso, line 91: Should we set up a redirect so someone visiting `/guides/seo` ends up in the correct place?

cprecioso (approved):
LGTM, just two loose ends

opensaas-sh/app_diff/main.wasp.diff
vincanger: Yeah this was. On the landing page of opensaas.sh `pricing` is behind the login as an example. I don't want pricing to be part of the OpenSaaS messaging anyway so that users don't get confused and think Open SaaS is a paid boilerplate.

opensaas-sh/blog/astro.config.mjs
vincanger: I don't think it's necessary. I don't think there will be any dead links out there, do you? What's your concern?

cprecioso: Yeah, my concern is dead links (maybe from an agent with older data?). I leave it up to you.

PR #679, opened 2026-05-18
Add llms.txt and JSON-LD schema markup to opensaas.sh

vincanger:
## Summary
- Generates `llms.txt` at build for AI crawlers
- Adds JSON-LD `SoftwareApplication`/`WebSite`/`Organization`/`FAQPage` schema to `<head>`

## Lighthouse (opensaas.sh, before)
- Agentic Browsing: **33** — fails `llms.txt is missing or incomplete`
- SEO: 92

Both issues addressed by this PR.


opensaas-sh/app_diff/scripts/generate-llms-txt.mjs.diff
cprecioso, line 1: Before I look into this script in depth, I think I want to propose an alternative.

This script seems very similar to the one in Wasp website, plus it also feels to me that more people would like to have this kind of functionality in their OpenSaaS apps (esp. in the Astro side)

Maybe it could be worth it to add this as an actual feature of OpenSaaS, instead of an ad-hoc script just for us? e.g. a bundled script (or a new package we publish) that takes in an index skeleton that everyone can customize, and does the thing.

(Also, I see that there are already Astro plugins that generate the llms.txt, can we leverage those? very shallow search: https://github.com/4hse/astro-llms-txt)

opensaas-sh/app_diff/main.wasp.diff
```diff
@@ -44,11 +44,13 @@
...
...ps://discord.gg/aCamt5wCpS\"]&#125;,&#123;\"@type\":\"FAQPage\",\"@id\":\"https://opensaas.sh/#faq\",\"mainEntity\":[&#123;\"@type\":\"Question\",\"name\":\"Why is this SaaS Template free and open-source?\",\"acceptedAnswer\":&#123;\"@type\":\"Answer\",\"text\":\"We believe the best product is made when the community puts their heads together. We also believe a quality starting point for a web app should be free and available to everyone. Our hope is that together we can create the best SaaS template out there and bring our ideas to customers quickly.\"&#125;&#125;,&#123;\"@type\":\"Question\",\"name\":\"What's Wasp?\",\"acceptedAnswer\":&#123;\"@type\":\"Answer\",\"text\":\"It's the fastest way to develop full-stack React + NodeJS + Prisma apps and it's what gives this template superpowers. Wasp relies on React, NodeJS, and Prisma to define web components and server queries and actions. Wasp's secret sauce is its compiler which takes the client, server code, and config file and outputs the client app, server app and deployment code, supercharging the development experience. Combined with this template, you can build a SaaS app in record time.\"&#125;&#125;]&#125;]&#125;</script>",
```
cprecioso: This feels like overloading the page too much. Since React 19 you can add `<script>` tags in the React component proper, and with prerendering they'll become part of the HTML that gets served to bots. Does that make sense to do?

template/app/src/client/components/DarkModeSwitcher.tsx
```diff
@@ -16,6 +16,8 @@ export function DarkModeSwitcher() {
       >
         <input
           type="checkbox"
+          aria-label="Toggle dark mode"
+          checked={!isInLightMode}
```
cprecioso, line 20: Cool

opensaas-sh/app_diff/scripts/generate-llms-txt.mjs.diff
vincanger: So these plugins all require us to upgrade Astro/Starlight. So I just added a boilerplate llms.txt to the app's `/public` dir and docs with info on filling it in, and once we take care of the astro upgrade, we can add the plugin: https://github.com/wasp-lang/open-saas/issues/682

opensaas-sh/app_diff/main.wasp.diff
vincanger: ah nice. fixed! I confirmed it worked with `wasp build start` too.

template/app/src/landing-page/components/SchemaMarkup.tsx
cprecioso, line 1: Really nice!

cprecioso (approved):
ship it

PR #687, opened 2026-05-25
Migrate opensaas template and opensaas-sh to the new TS spec

FranjoMindek:
## Description

Besides the DSL -> TS spec migration, I also:
- replace the `axios` call with `ky` (in line with Wasp's migration from `axios` to `ky`)

## Contributor Checklist


opensaas-sh/app_diff/src/file-upload/file-upload.wasp.ts.diff
```diff
@@ -0,0 +1,28 @@
...
+   action(createFileUploadUrl, { entities: ["User", "File"] }),
+   action(addFileToDb, { entities: ["User", "File"] }),
+   action(deleteFile, { entities: ["User", "File"] }),
++  job(deleteFilesJob, {
++    executor: "PgBoss",
++    schedule: {
++      cron: "0 5 * * *", // every day at 5am
++    },
++    entities: ["File"],
++  }),
```
FranjoMindek, line 27: Maybe we should include this as part of the template?

opensaas-sh/app_diff/src/file-upload/fileUploading.ts.diff
```diff
@@ -2,12 +2,12 @@
 +++ opensaas-sh/app/src/file-upload/fileUploading.ts
 @@ -1,4 +1,6 @@
 +import { PrismaClient } from "@prisma/client";
- import axios from "axios";
```
FranjoMindek, line 5: `axios` was never declared as a dependency, it used the Wasp's `axios` instance. Now we have to use `ky`.

template/app/src/admin/dashboards/messages/MessagesPage.tsx
```diff
@@ -2,7 +2,7 @@
 import type { AuthUser } from "wasp/auth";
 import { DefaultLayout } from "../../layout/DefaultLayout";
 
-export function AdminMessages({ user }: { user: AuthUser }) {
```
FranjoMindek, line 5: In line with other pages in this "domain".

template/e2e-tests/playwright.config.ts
```diff
@@ -39,7 +39,7 @@ export default defineConfig({
    * For local development, where this does pose a nuisance, we start the app / web server manually with `wasp db start` and `wasp start` and then start tests with `npm run local:e2e:start`.
    */
   webServer: {
-    command: "run-wasp-app dev --path-to-app=../app --wasp-cli-cmd=wasp",
```
FranjoMindek, line 42: Should `wasp-app-runner` check for this and run `wasp install` as necessary?
Most likely yes?

This is a temporary measure then.

template/app/src/admin/admin.wasp.ts
```diff
@@ -0,0 +1,43 @@
+import { page, route, type Part } from "@wasp.sh/spec";
+
+import { AnalyticsDashboardPage } from "./dashboards/analytics/AnalyticsDashboardPage" with { type: "ref" };
+import { MessagesPage } from "./dashboards/messages/MessagesPage" with { type: "ref" };
+import { UsersDashboardPage } from "./dashboards/users/UsersDashboardPage" with { type: "ref" };
+import { CalendarPage } from "./elements/calendar/CalendarPage" with { type: "ref" };
+import { SettingsPage } from "./elements/settings/SettingsPage" with { type: "ref" };
+import { ButtonsPage } from "./elements/ui-elements/ButtonsPage" with { type: "ref" };
+
+export const adminParts: Part[] = [
```
sodic: This might be a nice opportunity to showcase user-land helpers

template/app/src/auth/auth.wasp.ts
```diff
@@ -0,0 +1,74 @@
...
+    // Uncomment to enable GitHub Auth (check https://wasp.sh/docs/auth/social-auth/github for setup instructions):
+    // gitHub: {
+    //   userSignupFields: getGitHubUserFields,
+    //   configFn: getGitHubAuthConfig,
+    // },
+    // Uncomment to enable Discord Auth (check https://wasp.sh/docs/auth/social-auth/discord for setup instructions):
+    // discord: {
+    //   userSignupFields: getDiscordUserFields,
+    //   configFn: getDiscordAuthConfig,
+    // },
```
sodic: Might make sense to take them out as variables now that we can.

template/app/main.wasp.ts
```diff
@@ -0,0 +1,84 @@
...
+      email: "me@example.com",
+    },
+  },
+  parts: [
+    // Prerendering routes with static content creates HTML files at build time that are served immediately,
+    // improving SEO, search engine/AI crawling, and performance: https://wasp.sh/docs/advanced/prerendering
+    route("LandingPageRoute", "/", page(LandingPage), { prerender: true }),
+    route("NotFoundRoute", "*", page(NotFoundPage)),
+
+    ...authParts,
```
sodic: Maybe drop the parts suffix like we do in kitchensink?

template/app/main.wasp.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import { App } from "./src/client/App" with { type: "ref" };
+import { NotFoundPage } from "./src/client/components/NotFoundPage" with { type: "ref" };
+import { serverEnvValidationSchema } from "./src/env" with { type: "ref" };
+import { LandingPage } from "./src/landing-page/LandingPage" with { type: "ref" };
+import { seedMockUsers } from "./src/server/scripts/dbSeeds" with { type: "ref" };
+
+import { adminParts } from "./src/admin/admin.wasp";
+import { analyticsParts } from "./src/analytics/analytics.wasp";
+import { auth, authParts } from "./src/auth/auth.wasp";
```
sodic: Hmm, strange.

template/app/main.wasp.ts
```diff
@@ -0,0 +1,84 @@
...
+import { demoAiAppParts } from "./src/demo-ai-app/demo-ai-app.wasp";
+import { fileUploadParts } from "./src/file-upload/file-upload.wasp";
+import { paymentParts } from "./src/payment/payment.wasp";
+import { userParts } from "./src/user/user.wasp";
+
+export default app({
```
sodic, line 19: Maybe parts of this would also benefit from extraction into variables?

template/app/src/file-upload/file-upload.wasp.ts
```diff
@@ -0,0 +1,23 @@
...
+  route(
+    "FileUploadRoute",
+    "/file-upload",
+    page(FileUploadPage, { authRequired: true }),
+  ),
+  query(getAllFilesByUser, { entities: ["User", "File"] }),
+  query(getDownloadFileSignedURL, { entities: ["User", "File"] }),
+  action(createFileUploadUrl, { entities: ["User", "File"] }),
+  action(addFileToDb, { entities: ["User", "File"] }),
+  action(deleteFile, { entities: ["User", "File"] }),
```
sodic, line 22: Looks like a textbook usecase for a helper.

Just an idea from something I tried the other day:

Image

sodic (commented):
Left some comments. Let's use the new capabilities to write the best spec we can for opensaas.

template/app/main.wasp.ts
FranjoMindek: We need to agree on one naming schema.
We have:

<table>
<tr>
 <td>starters
 <td>examples
<tr>
 <td>`auth`
 <td>`authConfig`
<tr>
 <td>`authParts`
 <td>`auth`
<tr>
 <td>`domainParts`
 <td>`domain`
</table>

We should apply one to all.

template/app/src/auth/auth.wasp.ts
FranjoMindek: It would be cool if we could do this (not related to variables):

```ts
google:
  process.env.GOOGLE_CLIENT_ID && process.env.GOOGLE_CLIENT_SECRET
    ? {
        userSignupFields: getGoogleUserFields,
        configFn: getGoogleAuthConfig,
      }
    : undefined,
```

Should we have it as an userland feature (we actually don't want to provide these to the TS spec due to safety... etc.) , or should we implement this ourselves?

FranjoMindek: > Might make sense to take them out as variables now that we can.

You mean as in:
Define them as "unused variables", and if they want to use it they can just plug it into the config?

template/app/main.wasp.ts
FranjoMindek: I thought we preferred to keep it as a single object if possible.
I.e. extract only if we have to re-use some route.

template/app/src/admin/admin.wasp.ts
FranjoMindek: I did this.
Should showcase how to easily apply custom rules.

```ts
export const adminParts: Part[] = [
  createAdminRoute("AdminRoute", "/admin", AnalyticsDashboardPage),
  createAdminRoute("AdminUsersRoute", "/admin/users", UsersDashboardPage),
  createAdminRoute("AdminSettingsRoute", "/admin/settings", SettingsPage),
  createAdminRoute("AdminCalendarRoute", "/admin/calendar", CalendarPage),
  createAdminRoute("AdminUIButtonsRoute", "/admin/ui/buttons", ButtonsPage),
  // TODO: Add functionality to allow users to send messages to admin
  //       and make them accessible via the admin dashboard.
  createAdminRoute("AdminMessagesRoute", "/admin/messages", MessagesPage),
];

/**
 * All admin routes require authentication.
 */
function createAdminRoute(
  name: Route["name"],
  path: Route["path"],
  pageComponent: Page["component"],
): Route {
  return route(name, path, page(pageComponent, { authRequired: true }));
}
```

Did try to see if this would work, but no, going through constructor we cant:
image

image

Just some thought experiments.

template/app/src/file-upload/file-upload.wasp.ts
FranjoMindek: I did this, with param names different.

I'm think what's the best palce to leave this helepr at.
Currently I use it at two different places, so I left it at `main.wasp.ts`.
But honestly, even `helpers.wasp.ts` at root would work?

template/app/src/admin/admin.wasp.ts
sodic: I don't understand, did this not work? Why not?

>  but no, going through constructor we cant:

What does this mean?

template/app/src/auth/auth.wasp.ts
sodic: > You mean as in:
> Define them as "unused variables", and if they want to use it they can just plug it into the config?

Yes, that's what I meant

> It would be cool if we could do this (not related to variables):

This I would not do. It flips the source of truth. Env variables should be in service of auth methods you want to have, not the other way around.

template/app/main.wasp.ts
sodic: What do you prefer?

Btw, keep in mind we'll rename parts to decls.

template/app/main.wasp.ts
sodic: For parts, I'd say yes, but other fields could maybe be nicer this way. I'll let @vincanger decide what looks best.

template/app/src/file-upload/file-upload.wasp.ts
sodic: Whatever you think is the most beginner friendly.

Don't forget that this is a template, it's meant to teach as much as it's meant to work.

template/app/src/admin/admin.wasp.ts
FranjoMindek: Look at the return type, it's a `Route` which as a `Page` with `true` on `authRequired`.

We don't do generics magic in constructors so it does not return the actual return type, but rather just `Route` or `Page`.


The constructors work fine with the `Route` return type.

template/app/src/file-upload/file-upload.wasp.ts
FranjoMindek: image
Can't be in root, but yes for `src/`.

FranjoMindek: Decided to opt out for `src/shared/helpers.wasp.ts` so users can know they can make `.wasp.ts` which do not expose "configuration".

(might change helpers to something else like utils)

PR #690, opened 2026-06-01
vinces-ts-spec-migration

vincanger:
## Summary

Mostly just separated the spec files by feture, with some cleaning up of each config section.

## Notes from TS Spec review

Here are some comments on the TS Spec while migrating Open SaaS:

**Migration prompt for docs** — add a copy-paste agent prompt (with correct docs URLs, version numbers, etc.) to help users migrate from the DSL to TS Spec.

**Version string enums** — expected the version string to have enums/type checking for better DX.
version string enums

**Hide `.wasp/out` import** — the internal `.wasp/out` import path probably shouldn't be surfaced to users.
hide wasp/out import

**Wasp logo on `.wasp.ts` files** — would be a nice DX touch to show the Wasp logo on `*.wasp.ts` files in the editor.
wasp logo on .wasp.ts files


template/app/main.wasp.ts
```diff
@@ -0,0 +1,45 @@
...
+  auth: authConfig, 
+  emailSender: emailSenderConfig,
+  client: { rootComponent: App },
+  server: { envValidationSchema: serverEnvValidationSchema },
+  db: {
+    // Run `wasp db seed` to seed the database with the seed functions below:
+    seeds: [seedMockUsers], // Populates the database with a bunch of fake users to work with during development.
+  },
+  parts: [
+    // Prerendering creates static HTML at build time: https://wasp.sh/docs/advanced/prerendering
```
FranjoMindek: Probably move SEO comment here? Since we prerender here too?

FranjoMindek:
Thanks for the input vinny, closing now.

PR #703, opened 2026-07-03
fix(#688): Add Stripe Webhook Idempotency check via StripeEvent entity

### LINKED ISSUE #688: Missing idempotency in webhook endpoints (+ related robustness gaps)
## Summary

A review of the Stripe webhook handler (`src/payment/stripe/webhook.ts`) against Stripe's official [webhook best practices](https://docs.stripe.com/webhooks.md) and [signature verification](https://docs.stripe.com/webhooks/signature.md) docs surfaced a few correctness/robustness gaps. The most important is the lack of **idempotency**, which can double-count purchased credits.

### What's already correct ✅

- Raw request body is preserved via `stripeMiddlewareConfigFn` (`express.json` removed, `express.raw` installed) — the most common webhook bug, handled correctly.
- Signature verification via `stripeClient.webhooks.constructEvent(rawBody, signature, secret)`; missing signature header throws → `400`.
- Explicit event-type allow-list; unhandled events return `2xx` so Stripe doesn't retry them indefinitely.
- Replay/timestamp tolerance handled by the SDK default.

---

### 🔴 High — No idempotency; duplicate `invoice.paid` double-counts credits

Stripe delivers events **at-least-once** and retries on any non-2xx response. `updateUserCredits` increments credits (`credits: { increment: numOfCreditsPurchased }` in `src/payment/user.ts`), so a redelivered `invoice.paid` event for the `Credits10` plan grants the credits **again**. Subscription updates are naturally idempotent (they set absolute values), but credit grants are not.

**Suggested fix:** Persist each processed `event.id` (e.g. a `ProcessedStripeEvent` model with a unique constraint) and short-circuit if already seen, before performing any mutation. See "Handle duplicate events" in the Stripe docs.

### 🟠 Medium — Cancellation email can be sent repeatedly

`customer.subscription.updated` fires frequently. The handler sends the cancellation ("We hate to see you go") email on **every** update where `cancel_at_period_end` is true — not just on the transition. Combined with the idempotency gap, users can receive duplicate emails. Gate it on the actual `false → true` transition (tracked in the DB) or behind event-id dedupe.

### 🟠 Medium — Synchronous processing (including email send) before returning 2xx

Stripe recommends returning `2xx` immediately and doing heavy work asynchronously, because a slow handler can time out → Stripe retries → duplicate processing. The handler currently does DB writes and an `emailSender.send` network call inside the request path before responding. Consider offloading to a Wasp job. (Acceptable at low volume; flagging as a scaling concern.)

### 🟡 Low — Deterministic business errors return 400 → repeated pointless retries

The catch-all returns `400` for any `Error`, which is correct for signature/transient failures (a retry is desirable) but also applies to deterministic failures like "There should be exactly one line item" / "Unable to extract price id". Stripe will keep retrying those even though they can never succeed. Consider distinguishing transient errors (`400`, retry) from permanent business-logic errors (log + `2xx`, no retry) — similar to how `UnhandledWebhookEventError` is already handled.

### 🟡 Low — Event ordering assumed

Subscription handlers act on the event payload snapshot. Stripe does not guarantee ordering, so a stale `customer.subscription.updated` could overwrite newer state. The docs suggest fetching the latest object from the API when ordering matters. Likely fine given the current status-mapping logic, but noted.

---

## Suggested priority

1. Add event-id idempotency (fixes the High + first Medium at once).
2. Distinguish retry vs. no-retry per error class.
3. Consider async processing as volume grows.

## References

- https://docs.stripe.com/webhooks.md
- https://docs.stripe.com/webhooks/signature.md

cekuu35:
### Summary
This PR resolves the critical vulnerability described in #688 where the lack of idempotency handling in the Stripe webhook handler could cause a redelivered `invoice.paid` event to double-count purchased credits.

### Changes Made
1. **Added `StripeEvent` Model:** Introduced a simple schema entity to store successfully processed Stripe Event IDs (`evt_...`).
2. **Updated `main.wasp.ts`:** Exposed the `StripeEvent` entity to the `paymentsWebhook` API route inside `payment.wasp.ts`.
3. **Idempotency Logic in `webhook.ts`:**
   - Performs a fast look-up check for `event.id` before handling any database updates or network requests (like email sending).
   - Gracefully handles concurrency / race conditions by catching unique constraint collisions (Prisma `P2002`) and responding with `204 No Content`.

### Verification & Testing
- Simulated duplicate event delivery via Stripe CLI by sending the same `invoice.payment_succeeded` payload twice.
- The first event was processed, user credits were credited, and `StripeEvent` record was created.
- The second event short-circuited with a `204 No Content` status and skipped updating credits, confirming no double-credit counting occurs.

Fixes #688

### DIFF
--- a/template/app/schema.prisma
+++ b/template/app/schema.prisma
@@ -109,3 +109,9 @@ model ContactFormMessage {
   isRead                    Boolean         @default(false)
   repliedAt                 DateTime?
 }
+
+model StripeEvent {
+  id        String   @id
+  createdAt DateTime @default(now())
+}
+
--- a/template/app/src/payment/payment.wasp.ts
+++ b/template/app/src/payment/payment.wasp.ts
@@ -21,7 +21,7 @@ export const paymentSpec: Spec = [
   query(getCustomerPortalUrl, { entities: ["User"] }),
   action(generateCheckoutSession, { entities: ["User"] }),
   api("POST", "/payments-webhook", paymentsWebhook, {
-    entities: ["User"],
+    entities: ["User", "StripeEvent"],
     middlewareConfigFn: paymentsMiddlewareConfigFn,
   }),
 ];
--- a/template/app/src/payment/stripe/webhook.ts
+++ b/template/app/src/payment/stripe/webhook.ts
@@ -34,6 +34,27 @@ export const stripeWebhook: PaymentsWebhook = async (
   try {
     const event = constructStripeEvent(request);
 
+    // Idempotency check: Prevent duplicate processing of the same Stripe event
+    const eventId = event.id;
+    try {
+      const alreadyProcessed = await context.entities.StripeEvent.findUnique({
+        where: { id: eventId },
+      });
+      if (alreadyProcessed) {
+        console.info(`[Webhook] Event ${eventId} already processed. Skipping...`);
+        return response.status(204).send();
+      }
+      await context.entities.StripeEvent.create({
+        data: { id: eventId },
+      });
+    } catch (err: any) {
+      if (err.code === 'P2002') {
+        console.info(`[Webhook] Event ${eventId} processed concurrently. Skipping...`);
+        return response.status(204).send();
+      }
+      throw err;
+    }
+
     // If you'd like to handle more events, you can add more cases below.
     // When deploying your app, you configure your webhook in the Stripe dashboard
     // to only send the events that you're handling above.


FranjoMindek:
Hi @cekuu35 , while this is a simple fix for idempotency problem, we should implement it for all payment processors rather than just Stripe. I know the original issue only mentioned Stripe, but the problem lies in all of them (since all of them depend on webhooks).

PR #704, opened 2026-07-08
Fix Wasp wrapper cache busting

infomiho:
## Summary
- Resolve pkg.pr.new current published Wasp main commit from the `x-commit-key` response header
- Run the commit-pinned pkg.pr.new package instead of mutable `@main`, so npm cache keys update when pkg.pr.new updates
- Keep a clear failure message if pkg.pr.new does not expose the published commit

## Testing
- `./tools/wasp version`
- `shellcheck tools/wasp`

### DIFF
--- a/tools/wasp
+++ b/tools/wasp
@@ -18,4 +18,4 @@ if [ -z "$latest_commit" ]; then
   exit 1
 fi
 
-npm_config_prefer_online=true npx -y "https://pkg.pr.new/@wasp.sh/wasp-cli@$latest_commit" "$@"
+exec npx -y "https://pkg.pr.new/@wasp.sh/wasp-cli@$latest_commit" "$@"
--- a/tools/wasp
+++ b/tools/wasp
@@ -2,4 +2,11 @@
 
 # Just a wrapper to always use the latest Wasp CLI from main branch.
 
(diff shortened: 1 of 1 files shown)


tools/wasp
```diff
@@ -2,4 +2,11 @@
...
 
-npm_config_prefer_online=true npx -y https://pkg.pr.new/@wasp.sh/wasp-cli@main "$@"
+latest_commit=$(curl -fsLI https://pkg.pr.new/@wasp.sh/wasp-cli@main | grep -i '^x-commit-key:' | cut -d: -f4- | tr -d '[:space:]')
+
+if [ -z "$latest_commit" ]; then
+  printf '%s\n' "Failed to resolve latest published Wasp commit from main." >&2
+  exit 1
+fi
+
+npm_config_prefer_online=true npx -y "https://pkg.pr.new/@wasp.sh/wasp-cli@$latest_commit" "$@"
```
cprecioso: We can now drop the prefer_online thing.
Edit: And we can also use `exec` so that the bash process doesn't keep active when we call npx:


```suggestion
exec npx -y "https://pkg.pr.new/@wasp.sh/wasp-cli@$latest_commit" "$@"
```

cprecioso (approved):
yes

